PR #9, opened 2021-11-16
Add questions file support to enable user form in Rancher

juadk:
Fix https://github.com/epinio/epinio/issues/976

This PR mainly includes the support of the questions file to enable user form to configure the Chart:
image

It also contains two minor updates about the official email address to use as well as new options for the Helm chart.

### DIFF
--- a/chart/epinio/questions.yml
+++ b/chart/epinio/questions.yml
@@ -1,0 +1,128 @@
+questions:
+- variable: email
+  label: Email
+  description: "Email to use for getting notifications about your certificates"
+  default: "epinio@suse.com"
+  type: strings
+  required: false
+  group: "General settings"
+- variable: tlsIssuer
+  label: Tls issuer
+  description: "Name of the cluster issue to use"
+  default: "epinio-ca"
+  type: enum
+  required: false
+  group: "General settings"
+  options:
+    - "epinio-ca"
+    - "selfsigned-issuer"
+    - "letsencrypt-production"
+- variable: user
+  label: API username
+  description: "The user name for authenticating all API requests"
+  type: strings
+  required: false
+  group: "General settings"
+- variable: password
+  label: API password
+  description: "The password for authenticating all API requests"
+  type: password
+  required: false
+  group: "General settings"
+- variable: domain
+  label: Domain
+  description: "Domain for the application"
+  type: strings
+  required: false
+  group: "General settings"
+- variable: accessControlAllowOrigin
+  label: Access control allow organization
+  description: "Domain allowed to access the API. Comma-separated."
+  type: strings
+  required: false
+  group: "General settings"
+- variable: forceKubeInternalRegistryTls
+  label: Force tls for internal registry
+  required: false
+  type: boolean
+  group: "General settings"
+- variable: skipCertManager
+  label: Skip Cert Manager installation
+  required: false
+  type: boolean
+  group: "General settings"
+- variable: skipLinkerd
+  label: Skip Linkerd installation
+  required: false
+  type: boolean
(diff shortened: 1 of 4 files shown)


juadk:
### How to test the new form
You can provide the infra in many other ways but this is how I test it:

1. Deploy rancher under docker
`docker run -d --restart=unless-stopped -p 80:80 -p 443:443 --privileged rancher/rancher:v2.6.2`

2. Deploy a k3s cluster
I'm using vagrant to quickly pop a k3s cluster on my laptop with libvirt

3. Import the k3s cluster in Rancher
4. Add the custom helm chart repo `https://juadk.github.io/epinio-helm-chart`
image

5. Use the Helm chart to deploy Epinio
image
Click on the install button:
image

6. Customize by filling info in the form, once done, click on the install button and Epinio will be deployed in your cluster according your configuration

For now, I didn't test all the options, I only did a basic test with API user / password customization.

chart/epinio/questions.yml
```diff
@@ -0,0 +1,128 @@
...
+  group: "General settings"
+- variable: domain
+  label: Domain
+  description: "Domain for the application"
+  type: strings
+  required: false
+  group: "General settings"
+- variable: accessControlAllowOrigin
+  label: Access control allow organization
+  description: "Domain allowed to access the API. Comma-separated."
```
richard-cox, line 40: - `Domain allowed to access the API` - this is true, however it would be good to ensure the user knows they must configure this for the UI to work. I think the questions framework may support tooltips, if so could we add something like `Should match the domain which servers the UI`?
- `Comma-separated` - This isn't quite supported, I think Epinio places this value directly into the response header - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#directives

chart/epinio/values.yaml
```diff
@@ -2,10 +2,10 @@
 ## This is a YAML-formatted file.
 ## Declare variables to be passed into your templates.
 
-# The system domain to use for Epinio. Equivalent to `epinio install` system-domain
-# parameter.
+# Domain allowed to access the API. Comma-separated
```
richard-cox, line 5: This would need updating following any changes to the form

chart/epinio/questions.yml
juadk: Regarding tooltips, I don't see anything about it in the documentation https://rancher.com/docs/rancher/v2.6/en/helm-charts/creating-apps/#additional-files-for-rancher-charts
But, I can simply improve the description field.

richard-cox: Ahh, my mistake, they're not there (see `rancher/dashboard` `components/Questions/String.vue`). 

How about something like `Domain which serves the UI (to access the API)`?

juadk: I'm ok for `Domain which serves the Rancher UI (to access the API)` if you agree too.

juadk:
Just tested with filling the field `Access control allow origin` and it worked well, Rancher was able to communicate with the Epinio API.

chart/epinio/questions.yml
richard-cox: Looking good! Can't comment on the rest but this bit is fine

juadk:
Just to mention that I did more test with epinio 0.2.0 and the helm chart installation.
- Installation on K3s and GKE with default option :heavy_check_mark: 
- Installation with domain + external registry (docker hub) on GKE :heavy_check_mark: 
- Installation with domain + S3 on GKE :heavy_check_mark: 
- Uninstallation for every cases mentioned above :heavy_check_mark:

PR #12, opened 2021-11-24
Remove skip namespace option and set system-domain as mandatory option

juadk:
`--skip-default-namespace` option was removed from epinio (https://github.com/epinio/epinio/pull/983/commits/b31fdcdac310919714eb48a7b2ae33c3e9a576a5)
`--system-domain` is now a mandatory option (https://github.com/epinio/epinio/pull/983/commits/f770c645968e24be836c8d734c2c79bec4a92bc0)


ldevulder (approved):
lgtm

PR #14, opened 2021-12-02
Add the Epinio "small" chart

jimmykarily:
https://github.com/epinio/epinio/issues/1028


juadk:
I removed the installation part done by chart testing.We will configure it in another place later.

PR #24, opened 2021-12-08
Remove automated release of Helm chart

juadk:
The idea is to trigger the release manually with workflow_dispatch (will be introduced by another PR later).
Let's quickly remove the automation to don't mess up the project.


.github/workflows/release.yml
```diff
@@ -1,50 +0,0 @@
-name: Release Charts
-
-on:
-  push:
```
manno, line 4: Isn't it sufficient to remove the "on: push:" part?

jimmykarily: I thought we only wanted to change the trigger. We still want to release. Maybe by running a modified version of this pipeline with a manual trigger?

juadk: No it isn't sufficient, because if we receive the hook from epinio, it will do a release as well. (repository_dispatch)
I will open a new PR to add the automation about PR creation when we release Epinio, still using repository_dispatch.

juadk: > I thought we only wanted to change the trigger. We still want to release. Maybe by running a modified version of this pipeline with a manual trigger?

Yes, that's my plan, I though to do it in another PR.

juadk: I though you were blocked by this automation... Otherwise I can do all the changes in this PR.

jimmykarily (approved):

.github/workflows/release.yml
jimmykarily: sounds good then

manno: Ah, we want to introduce a PR here as an intermediate step? Ok.

A manual dispatch to trigger -prerelease versions would be nice, but not urgent. The AKS/EKS pipeline did something like that.

juadk: @manno Yes, that's the idea to use `workflow_dispatch` to manually start the release process.

PR #30, opened 2021-12-09
Fix cert-manager issue

ldevulder:
We should not rely on cert-manager if it is not installed during Epinio deployment.

### DIFF
--- a/chart/epinio-installer/templates/manifest.yaml
+++ b/chart/epinio-installer/templates/manifest.yaml
@@ -202,8 +202,9 @@ stringData:
             type: annotation
 
       - id: registry
-        # needs: namespace-registry
+        {{- if not .Values.skipCertManager }}
         needs: cert-manager
+        {{- end }}
         namespace: epinio-registry
         type: helm
         source:


chart/epinio-installer/templates/manifest.yaml
```diff
@@ -202,8 +202,11 @@ stringData:
             type: annotation
 
       - id: registry
-        # needs: namespace-registry
+        {{- if not .Values.skipCertManager }}
         needs: cert-manager
+        {{- else }}
+        needs: namespace-registry
```
manno, line 208: The helm chart will create that namespace. I think  `needs: namespace-registry` is only needed if we're using linkerd and want the namespace component to create the annotations.

ldevulder: Yes, according to this code
```
      - id: namespace-registry
        {{- if not .Values.skipLinkerd }}
        needs: linkerd
        {{- end }}
        type: namespace
        namespace: epinio-registry
        values:
          - name: linkerd.io/inject
            value: enabled
            type: annotation
```
It seems that the `namespace-registry` creation is done apart only to add an annotation for linkerd (as you mentioned), but what happens if the namespace is already created before by `epinio-registry`? Is there any failure in Helm installation? Or is the annotation is simply added? And the annotation is only needed for linkerd, so we may change the code like this, no?
```
      {{- if not .Values.skipLinkerd }}
      - id: namespace-registry
        needs: linkerd
        type: namespace
        namespace: epinio-registry
        values:
          - name: linkerd.io/inject
            value: enabled
            type: annotation
      {{- end }}
```

In fact my main concern here is when cert-manager is already installed: can we start the installation of `epinio-registry` at the same time as linkerd or not? (in case we don't have to install linkerd it's easier: it can be installed from the beginning).

ldevulder: Answer to myself after a test: the annotation is added even if the `epinio-registry` namespace is created before.

So as you suggested @manno the `needs: namespace-registry` can be safely removed.

PR #31, opened 2021-12-09
Improve Helm Chart CI pipeline

juadk:
This PR improves the CI pipeline:

-  When a pull request is opened
    1. Linter checks code as well as if the chart version is bumped
    2. If 1 is ok, it tests the installation (on top of k3d) of the epinio-installer chart (in any case, we always test only that chart, because it includes the others) 

- When we want to release Helm Charts
   1. Same steps as above without checking if versions are bumped (using specific `ct-release.yaml`)
   2. If 1 & 2 ok, we can execute chart releaser to publish the changes


.github/workflows/main-ci.yml
```diff
@@ -0,0 +1,51 @@
...
+          fi
+      - name: Set up k3d cluster
+        uses: nolar/setup-k3d-k3s@v1
+        with:
+          version: v1.22
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+        if: steps.list-changed.outputs.changed == 'true'
+
+      - name: Run chart-testing (install)
+        run: ct install --config ct/ct-install.yaml
```
jimmykarily, line 51: can all 3 of them be installed like this? I think the `epinio` one (the "small) can't be installed alone.

jimmykarily: oh I just saw the `ct-install.yaml` file now. We only install the big one right?

jimmykarily (approved):

.github/workflows/main-ci.yml
```diff
@@ -0,0 +1,51 @@
+name: Lint & install charts
+
+on: pull_request
```
olblak, line 3: Wouldn't it be better to also run this workflow "on main" and remove this part from the release process?
I think it's a bit redundant and that would save some release time.

.github/workflows/main-ci.yml
```diff
@@ -0,0 +1,51 @@
+name: Lint & install charts
```
olblak, line 1:
```suggestion
name: Lint and test chart installation
```
Would this clarify the meaning of this workflow displayed in pull request checks

chart/epinio-installer/ci/k3d-values.yaml
```diff
@@ -0,0 +1,60 @@
+## Default values for Epinio Helm Chart.
```
olblak, line 1: Is this file a replacement of the default chart  values.yaml because then I would just specify what need to be overridden.
I am mainly concerned by the following line parameters

```
containerRegistryChart: "https://github.com/epinio/epinio-helm-chart/releases/download/container-registry-0.1.0/container-registry-0.1.0.tgz"
epinioChart: "https://github.com/epinio/epinio-helm-chart/releases/download/epinio-0.2.0/epinio-0.2.0.tgz"
traefikChart: "https://helm.traefik.io/traefik/traefik-10.3.4.tgz"
``` 

I am wondering if this file should only contain

```
skipTraefik: true
```

.github/workflows/main-ci.yml
juadk: Yes you are right, only the big one :smile:

.github/workflows/main-ci.yml
juadk: We discussed a lot with @ldevulder about that yesterday and we thought it was good to test again lint and installation before releasing. We agree tests are done for every PR so the release will go fine. We see it as a double check :smiley: 
@jimmykarily WDYT?

chart/epinio-installer/ci/k3d-values.yaml
juadk: Yes, that's just a file to customize the test we are doing on top of k3d. 
https://github.com/helm/charts/blob/master/test/README.md#providing-custom-test-values
It can be useful if we want to test deployments with different parameters.
Maybe it can be improved with template but for now, I don't know, purpose is to have a quick CI to start.

manno (changes requested):
Is there a reason to have the CI files in the chart's dir? `helm package` will include those files in the tarball.
I'd propose to move ci to the top level.

juadk:
> Is there a reason to have the CI files in the chart's dir? `helm package` will include those files in the tarball. I'd propose to move ci to the top level.

@manno Yes it's mandatory to have CI files for specifics charts in chart dir.
_In order to enable custom test values, create a directory ci in the chart's directory and add any number of *-values.yaml files to this directory. Only files with a suffix -values.yaml are considered. Instead of using the defaults, the chart is then installed and tested separately for each of these files using the --values flag._
https://github.com/helm/charts/blob/master/test/README.md#providing-custom-test-values

I'll push a `.helmignore` file to exclude the ci folder.

.github/workflows/main-ci.yml
jimmykarily: I would also prefer that we run this on `main` and skip it on releases. I believe the release should only start when tests are passing (that is, on a green `main`). There is still some double checking, because we run the tests on PRs and then on `main` after merge. In most cases, merging a green branch shouldn't break `main` but there is still a chance this could happen if the branch wasn't rebased on `main` when tests run. So by moving this check on `main`, we catch that case early and not when we release.

jimmykarily:
>  it tests the installation (on top of k3d) of the epinio-installer chart (in any case, we always test only that chart, because it includes the others)

When we change one of the other charts (other than the installer one), we will still test the installation of the installer chart. That chart indeed includes a version of the other charts. But in this case, that version is not the version with the changes we just made (e.g. in a PR). So those changes are never tested.

I just want to make sure I got this right. I'm not suggesting we fix it on this PR because that involves, packaging and publishing the changed helm charts on some public location. We can do this on future PR.

manno (approved):

.github/workflows/main-ci.yml
juadk: Ok let me change that.

juadk:
> I just want to make sure I got this right. I'm not suggesting we fix it on this PR because that involves, packaging and publishing the changed helm charts on some public location. We can do this on future PR.

You are totally right. I will check how other projects are doing.

chart/epinio-installer/ci/k3d-values.yaml
juadk: It works perfectly with only `skipTraefik: true` in the file, thanks for that good point.

PR #34, opened 2021-12-14
Some additional passthrough

agracey:
I think this might possibly help my inability to install. But more generally, there's some places where the key and it's usage didn't match up.

### DIFF
--- a/chart/epinio-installer/questions.yml
+++ b/chart/epinio-installer/questions.yml
@@ -61,12 +61,6 @@ questions:
   required: false
   type: boolean
   group: "General settings"
-- variable: traceLevel
-  label: Log Level
-  required: false
-  type: strings
-  default: "2"
-  group: "General settings"
 - variable: useExternalRegistry
   default: "false"
   description: "Use an external registry"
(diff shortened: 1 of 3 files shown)


agracey:
This at least allows the debugging to be passed through

chart/epinio-installer/questions.yml
```diff
@@ -52,12 +52,12 @@ questions:
   type: boolean
   group: "General settings"
 - variable: skipLinkerd
-  label: Skip Linkerd installation
```
jimmykarily, line 55: Epinio can run even if there is no Linkerd. Maybe the new label is confusing in that case?

chart/epinio-installer/questions.yml
```diff
@@ -52,12 +52,12 @@ questions:
...
+  label: Use existing Linkerd Installation
   required: false
   type: boolean
   group: "General settings"
 - variable: skipTraefik
-  label: Skip Traefik installation
+  label: Use existing ingress controller
```
jimmykarily, line 60: Not sure if we can run on any other ingress controller yet. We need to try that. We have some Traefik specific annotations to enforce tls but they shouldn't be necessary.

jimmykarily (approved):

chart/epinio-installer/questions.yml
agracey: We should probably make those configurable through the chart as well?

chart/epinio-installer/questions.yml
agracey: Yeah, it could be. But I think it's closer to how a user looking at this with minimal background would expect to see it. We can document that if they don't want a service mesh, to just say they already have one and certain future features fail gracefully.

agracey: Resolving and adding an issue to spend some time thinking through the Rancher specific workflow assumptions.

chart/epinio-installer/questions.yml
agracey: Resolving for same reason as above. https://github.com/epinio/helm-charts/issues/36

PR #38, opened 2021-12-15
Make image repository configurable in the epinio chart

jimmykarily:
This way a private repository can be used for airgapped installations.
The builder image is still coming from dockerhub by default but at least
it can be specified in the application manifest or with as a "push"
argument.

Fixes: https://github.com/epinio/epinio/issues/979

### DIFF
--- a/chart/epinio/templates/_helpers.tpl
+++ b/chart/epinio/templates/_helpers.tpl
@@ -1,0 +1,5 @@
+{{- define "registry-url" -}}
+{{- if .Values.registryURL -}}
+{{ trimSuffix "/" .Values.registryURL }}/
+{{- end -}}
+{{- end -}}
--- a/chart/epinio/templates/_helpers.tpl
+++ b/chart/epinio/templates/_helpers.tpl
@@ -1,0 +1,5 @@
+{{- define "registry-url" -}}
+{{- if .Values.registryURL -}}
+{{- printf "%s/" .Values.registryURL -}}
+{{- end -}}
+{{- end -}}
--- a/chart/epinio/templates/_helpers.tpl
+++ b/chart/epinio/templates/_helpers.tpl
@@ -1,5 +1,0 @@
-{{- define "registry-url" -}}
-{{- if .Values.registryURL -}}
-{{- printf "%s/" .Values.registryURL -}}
-{{- end -}}
-{{- end -}}
--- a/chart/epinio/templates/server.yaml
+++ b/chart/epinio/templates/server.yaml
@@ -238,7 +238,7 @@ spec:
             - name: REGISTRY_CERTIFICATE_SECRET
               value: "{{ .Values.registry.certificateSecret }}"
             {{- end }}
-          image: "{{ template "registry-url" }}{{ .Values.image.epinio.repository}}:{{ or .Values.image.epinio.tag .Chart.AppVersion }}"
+          image: "{{ template "registry-url" . }}{{ .Values.image.epinio.repository}}:{{ or .Values.image.epinio.tag .Chart.AppVersion }}"
           livenessProbe:
             httpGet:
               path: /ready
--- a/chart/epinio/templates/server.yaml
+++ b/chart/epinio/templates/server.yaml
@@ -238,7 +238,7 @@ spec:
             - name: REGISTRY_CERTIFICATE_SECRET
               value: "{{ .Values.registry.certificateSecret }}"
             {{- end }}
-          image: splatform/epinio-server:{{ .Chart.AppVersion }}
+          image: "{{ template "registry-url" . }}{{ .Values.server.containerImage }}:{{ .Chart.AppVersion }}"
           livenessProbe:
             httpGet:
               path: /ready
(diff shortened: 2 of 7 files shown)


chart/epinio/templates/tekton-buildpacks-task.yaml
```diff
@@ -132,6 +132,7 @@ spec:
         privileged: true
 
     - name: create
+      # TODO: This makes air-gapped installation hard to use. Has to be specified on every push.
```
jimmykarily, line 135: Default builder is defined in code here: https://github.com/epinio/epinio/blob/main/internal/manifest/manifest.go#L18

chart/epinio/values.yaml
```diff
@@ -5,9 +5,19 @@
...
 
+# The URL of the container registry from where to pull container images
+# for the various created Pods. Don't confuse this registry with the "Epinio
+# registry" which is the one where Epinio stores the application images.
+registryURL: ""
+
 # The system domain that will be used to access the epinio API server
 systemDomain: ""
 
+tekton:
```
olblak, line 16: What do you think to use the following format:
```
image:
    epinio:
        registry: splatform/epinio-server
        tag: v0.2.1
    bash:
        registry: library/bash (or just bash)
        tag: 5.1.4
    aws-cli:
        registry: amazon/aws-cli
        tag: 2.0.52
    kubectl: 
        registry: lachlanevenson/k8s-kubectl
        tag: v1.22.2
```

It's a personal taste but I prefer to have a different field for the tag as it's going to change more often

chart/epinio/templates/_helpers.tpl
```diff
@@ -0,0 +1,5 @@
+{{- define "registry-url" -}}
```
olblak, line 1: I am probably missing something but what's the purpose of this template versus just using {{ .Values.registryURL }}

jimmykarily: Oops, currently none. I started lazy and [copied from a place where it had more logic](https://github.com/rancher/charts/blob/dev-v2.6/charts/longhorn/longhorn/1.1.100/templates/_helpers.tpl#L49-L55). Then I removed the logic...

I will simplify this.

chart/epinio/values.yaml
jimmykarily: Looks cleaner, yes. I'm not sure we should call the key `registry` though. It will either be the combination of a registry + an image (e.g. `ghcr.io/epinio/epinio-server`) or just the image (e.g. `splatform/epinio-server`). So the part that's always there is the image.

Maybe we should remove the top-level `registryURL` and let users specify each image repository separately?

olblak: I am sorry, I meant "repository" and not registry. I don't think being able to override the default registry in some situations would be useful at least at this stage but I already see the value of being able to override the default registry URL to let say a local proxy cache.

olblak:
```
image:
    epinio:
        repository: splatform/epinio-server
        tag: v0.2.1
    bash:
        repository: library/bash (or just bash)
        tag: 5.1.4
    aws-cli:
        repository: amazon/aws-cli
        tag: 2.0.52
    kubectl: 
        repository: lachlanevenson/k8s-kubectl
        tag: v1.22.2
```

jimmykarily: oh it makes more sense now :) . I agree. Let me change things.

chart/epinio/templates/server.yaml
```diff
@@ -238,7 +238,7 @@ spec:
             - name: REGISTRY_CERTIFICATE_SECRET
               value: "{{ .Values.registry.certificateSecret }}"
             {{- end }}
-          image: splatform/epinio-server:{{ .Chart.AppVersion }}
+          image: "{{ trimSuffix "/" .Values.registryURL }}/{{ .Values.image.epinio.repository}}:{{ or .Values.image.epinio.tag .Chart.AppVersion }}"
```
olblak, line 241: Does that mean that if registryURL is set to "" image will be 

```
    image: /splatform/epinio:v0.2.0
```
Can it start with a "/"

jimmykarily: I didn't think of that. I was trying to make it work both with and without a `/` on the `registryURL`. Maybe it needs a helper after all...

jimmykarily: Maybe: 

```
 {{ if registryURL }}{{ trimSuffix "/" .Values.registryURL }}/{{ end }}{{ .Values.image.epinio.repository}}
```

olblak: Yep I now see the value of the helper

jimmykarily: ok re-introduced the helper. Hopefully it's fine now.

olblak (approved):

jimmykarily:
I installed epinio with the epinio-installer helm chart. I then fetched the values.yaml with:

```
helm get values epinio -n epinio > epinio-values.yaml
```

I edited that file to add the new `image.epinio.repository` key and pointed that to another image. I then did a `helm upgrade --values epinio-values.yaml -n epinio epinio chart/epinio` and the epinio-server pod was updated to use that other image.

I tried with the `registryURL` but it doesn't update the pods. Not sure, why. Checking.

jimmykarily:
> I tried with the `registryURL` but it doesn't update the pods. Not sure, why. Checking.

The template needs the context to be passed in order to access `.Values`:

https://github.com/epinio/helm-charts/pull/38/commits/97b0fa1f506cbfb4abe99c1c5a1ce1bffb0a76d8

PR #39, opened 2021-12-16
When using external registry don't try to mount registry secret in tekton pod

jimmykarily:
By setting the registry.certificateSecret to non empty value, the epinio
helm chart sets the REGISTRY_CERTIFICATE_SECRET environment variable on the
epinio server Deployment. This results in epinio server trying to mount
the secret on the tekton pod. In the case of external registry, that
secret doesn't exist.

https://github.com/epinio/epinio/blob/247ef0a8304ab130a8395d383570c73a65245722/internal/api/v1/application/stage.go#L190-L195
https://github.com/epinio/epinio/blob/247ef0a8304ab130a8395d383570c73a65245722/internal/api/v1/application/stage.go#L374-L379


juadk (approved):
Tested successfully! thanks

PR #52, opened 2021-12-22
Build and publish charts locally for the CI

juadk:
This PR improves our CI pipeline by building Helm charts and publishing them locally in [Chartmuseum](https://chartmuseum.com/). 
That means we can test latest version of our Helm Charts without waiting official release in gh-pages.

In short, we host chartmuseum instance inside the GH runner and epinio-installer will fetch container-registry and epinio from there by using `http://host.k3d.internal:8090/charts/container-registry-0.1.0.tgz`.
As it's ephemeral and we don't care about versioning rules in CI, we always force version 0.1.0.

I tested the workflow inside one repo of mine: https://github.com/juadk/epinio-helm-chart/runs/4601154380?check_suite_focus=true

Address https://github.com/epinio/epinio/issues/1017


scripts/build-charts-ci.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+set -e
+
+# Install chartmuseum using the gofish package manager
+curl -fsSL https://raw.githubusercontent.com/fishworks/gofish/main/scripts/install.sh | bash
+gofish init
+gofish install chartmuseum
```
jimmykarily, line 8: Can't we do this instead? https://github.com/helm/chartmuseum#helm-chart

This script will have to assume there is a Kubernetes cluster available, but we always have a cluster when running CI jobs right? We just have to make sure we run this script when that cluster is up and running.

juadk: I already tried to install chartmuseum with its chart in k3d cluster yesterday and it didn't work very well. I didn't want to spend more time on it as it's an ephemeral stack. I can try again but it will take time.

scripts/build-charts-ci.sh
```diff
@@ -0,0 +1,53 @@
...
+# Push all our charts to the repo
+# We force version 0.1.0 because we don't care about versionning
+# We only care about testing here
+cd chart/
+helm cm-push --version "0.1.0" epinio-installer/ chartmuseum
+helm cm-push --version "0.1.0" epinio/ chartmuseum
+helm cm-push --version "0.1.0" container-registry/ chartmuseum
+
+# Update epinio-installer's values.yaml file for using charts hosted in our repo
+sed -i -e 's|containerRegistryChart:.*|containerRegistryChart: "http://host.k3d.internal:8090/charts/container-registry-0.1.0.tgz"|'\
```
jimmykarily, line 51: I'm not sure this host is resolvable from within the installer job. I think `resolv.conf` is inherited all the way down to pods but I'm not sure this will work.

scripts/build-charts-ci.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# Add our new ephemeral repo
+helm repo add chartmuseum http://localhost:8090
+helm repo update
+
+# Push all our charts to the repo
+# We force version 0.1.0 because we don't care about versionning
+# We only care about testing here
+cd chart/
+helm cm-push --version "0.1.0" epinio-installer/ chartmuseum
```
jimmykarily, line 46: Shouldn't there be some change in this PR to install from this location? (maybe it's already there and I can't see it?)

scripts/build-charts-ci.sh
jimmykarily: fair

scripts/build-charts-ci.sh
juadk: I think it works because in a previous test, I forgot `/chart/` in the repo url and I got this:
https://github.com/juadk/epinio-helm-chart/runs/4598040631?check_suite_focus=true#step:11:134
And now with the good url repo, it passes: https://github.com/juadk/epinio-helm-chart/runs/4601243506?check_suite_focus=true

scripts/build-charts-ci.sh
juadk: We install from this location thank to https://github.com/epinio/helm-charts/pull/52/files#diff-4b4a9fbed581194852da125a443f567a145dbcadd7a0ea64be79f50f8470ec51R51
Am I wrong?

juadk: Yes we don't need to push epinio-installer to chartmuseum as chart testing is building it from dir. Let me remove it.

jimmykarily (approved):

PR #75, opened 2022-01-25
Enable dependabot for github action

olblak:


olblak:
I guess no objection so I am merging it for now but in worst case we can still revert later on.

PR #86, opened 2022-02-04
Use installer docker image version

manno:


manno:
This is blocked by updatecli not supporting docker image as a source, yet :)
I'll remove the update for now.

PR #88, opened 2022-02-04
Trim -noasset from epinio-installer release name

olblak:
It appears that the Epinio installer docker image tag doesn't have the string "-noassets" while the latest Github Release does.
Updatecli has the ability to "transform" a value as described on the documentation [transformers](https://www.updatecli.io/docs/core/transformer/) 


manno:
The docker image is independent of the github release.

PR #96, opened 2022-02-09
Install minio as a subchart of the epinio chart

jimmykarily:
Using this chart: https://github.com/minio/minio/tree/master/helm/minio
which doesn't deploy an operator (which we didn't need anyway).
This removes one dependency from the installer manifest. One more step
closer to single-chart.

Relevant discussion: https://github.com/epinio/epinio/discussions/1190


chart/epinio/Chart.yaml
```diff
@@ -12,4 +19,4 @@ maintainers:
 name: epinio
 sources:
 - https://github.com/epinio/epinio
-version: 0.3.0
+version: 0.4.0
```
ldevulder, line 22: Done by `updatecli`, has to be removed. Sorry my first thought was false...

chart/epinio-installer/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: epinio-installer
-version: 0.4.0
+version: 0.5.0
```
ldevulder, line 3: Done by `updatecli`, has to be removed. Sorry my first thought was false...

jimmykarily: done

chart/epinio/Chart.yaml
jimmykarily: done

jimmykarily:
We can use cert-manager and this helm chart to fix this: https://github.com/epinio/helm-charts/pull/97/files#r806568901

manno (approved):

ldevulder (approved):
Testing successfully on my lab. Let's see how how it will work on Public Clouds after merging :-)

PR #97, opened 2022-02-11
fix: missing abort on error, and linkerd shutdown from init containers

andreas-kupries:
Ref: https://github.com/epinio/epinio/issues/1209
See also https://github.com/epinio/epinio/pull/1210

### DIFF
--- a/chart/epinio/templates/stage-scripts.yaml
+++ b/chart/epinio/templates/stage-scripts.yaml
@@ -37,6 +37,10 @@ data:
     # the detected format more explicit in the output.
     #
     echo By _ _ __ ___ _____ $(whoami) $(pwd)
+    if test ! -f "/workspace/source/${BLOBID}" ; then
+      echo Nothing to unpack
+      exit
+    fi
     mkdir /workspace/source/app
     (  cd /workspace/source/app
        ( echo Tar? ; tar -xvf  "../${BLOBID}" 2>&1 | sed -e 's/^/Tar: /' ) || \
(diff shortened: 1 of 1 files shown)


ldevulder:
I did a quick test but it seems that `curl` call is failing when used with `trap`:
```
🕞  [stage.workspace.wordpress.12f8d4c91b3d4405-7ptpc] unpack-blob /stage-support/unpack: line 35: curl: command not found 
2022/02/11 17:34:18 EpinioApiClient: "msg"="response is not StatusOK" "error"="Internal Server Error: Failed to stage" "status"=500
error pushing app to server: waiting for staging failed: All attempts fail:
#1: Internal Server Error: Failed to stage
```

ldevulder:
Anyway: I think that it could be a good idea to add a regression test for this case, WDYT? I can add one after this fix is merged.

andreas-kupries:
`curl` is not available in the init containers. I.e. not provided by their images. I rewrote things to not need curl in the init containers now. See also comment https://github.com/epinio/epinio/pull/1210#issuecomment-1038792103 and following.

This actually passes CI now.

I agree that we should have regression tests for the various kinds of possible stage failures (download, unpack, buildpack).

ldevulder (approved):

chart/epinio/templates/stage-scripts.yaml
```diff
@@ -62,8 +66,14 @@ data:
     # container gracefully. We use `|| true` in case linkerd is not deployed.  Further, it
     # is placed into a trap to ensure that it will always run, even for a staging failure.
     #
+    set -e
     trap "curl -X POST http://localhost:4191/shutdown || true" EXIT
     echo By _ _ __ ___ _____ $(whoami) $(pwd)
+    if test ! -d "/workspace/source/app" ; then
+      echo Nothing to build
+      sleep 60 # linkerd is a pain - If we exit to quickly, with the sidecar not ready our curl to shut it down does nothing, and then the sidecar comes up and prevents the pod from ending
```
jimmykarily, line 74: The only reason we still keep the staging Job meshed is because it may talk to minio to fetch the source blob (if external S3 is not used).

The helm chart [we are planning to use](https://github.com/epinio/helm-charts/pull/96) supports passing a secret with certs which should make it possible to use cert-manager to create a cert and encrypt communication with minio: https://github.com/minio/minio/blob/master/helm/minio/values.yaml#L110-L116

If that works, we can "unmesh" the staging job and get rid of this trick.

andreas-kupries: Yes, having this in the future is pretty much the only reason I am ok with my hack here.
Not as a hacky, more complicated would have been code waiting for the sidecar to be up ...
Basically repeat the `/shutdown` until it connects and succeeds.

PR #101, opened 2022-02-15
Epinio single helm chart

jimmykarily:
Part of: https://github.com/epinio/epinio/issues/1217

* use kubed as a sub-chart
* remove container-registry chart, include files in epinio chart
* move params to  global


chart/epinio/templates/stage-scripts.yaml
```diff
@@ -15,7 +15,8 @@ data:
     # internal s3 the chart has no information.  Therefore we cannot
     # use helm templating to insert these.
     echo By _ _ __ ___ _____ $(whoami) $(pwd)
-    aws --endpoint-url "${PROTOCOL}://${ENDPOINT}" s3 cp "s3://${BUCKET}/${BLOBID}" "/workspace/source/${BLOBID}"
+    cat /certs/ca.crt /certs/tls.crt > /tmp/ca-bundle.pem
```
jimmykarily, line 18: I need to check but this may fail with lets encrypt because cert-manager doesn't put a CA in the generated secret

manno: Might fail, but I only put the CA here, because it's required to list the complete certificate chain. Just using the cert could also work.

However, over at https://github.com/epinio/epinio/compare/1217-single-chart#diff-646aebb58b95700e7e8bb4d9357e5111baf5a4f85bf80f3f0549c61ebb807227R149 we might have a problem

jimmykarily: ok did some changes on the epinio side to mount the CA and cert only if the argument is set. Let's keep it as it is and see

enrichman (dismissed):

PR #105, opened 2022-02-21
Deploy, update, and remove applications via helm - chart

andreas-kupries:
Ref https://github.com/epinio/epinio/issues/1224

The chart used by https://github.com/epinio/epinio/pull/1225

See also https://github.com/epinio/application/pull/5

### DIFF
--- a/chart/application/Chart.yaml
+++ b/chart/application/Chart.yaml
@@ -1,0 +1,10 @@
+apiVersion: v2
+name: epinio-application
+description: The helm chart for epinio applications to be deployed by
+icon: https://raw.githubusercontent.com/epinio/helm-charts/main/assets/epinio.png
+home: https://github.com/epinio/epinio
+type: application
+version: 0.1.10
+maintainers:
+- email: team@epinio.io
+  name: SUSE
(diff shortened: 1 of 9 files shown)


andreas-kupries:
I do not know why the installation test thing fails. Help please.

chart/application/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: epinio-application
+description: The helm chart for epinio applications to be deployed by
+icon: https://raw.githubusercontent.com/epinio/helm-charts/main/assets/epinio.png
+home: https://github.com/epinio/epinio
+type: application
+version: 0.1.8
+appVersion: "1.16.0"
```
jimmykarily, line 8: Should these versions match something? What is `appVersion` in this case?

andreas-kupries: e69a351c48a6bd27093811423cadf4306bdc320c App version now matches version. Keep it matched in the future.

andreas-kupries:
We know the issue now. The CI attempts to install the chart, with default values. The app chart is an app framework, not a complete app. It will not install. 

@juadk has volunteered to look at how to exclude this chart from the install testing.

enrichman (approved):

jimmykarily (approved):

PR #111, opened 2022-02-28
Remove "without dependencies" message

juadk:
Remove "without dependencies" message out of the Epinio helm chart description

### DIFF
--- a/chart/epinio/Chart.yaml
+++ b/chart/epinio/Chart.yaml
@@ -13,7 +13,7 @@ dependencies:
   tags:
   - kubed
   version: 0.13.1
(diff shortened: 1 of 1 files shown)


chart/epinio/Chart.yaml
```diff
@@ -13,7 +13,7 @@ dependencies:
   tags:
   - kubed
   version: 0.13.1
-description: The Epinio component (without dependencies)
+description: The Epinio component
```
jimmykarily, line 16: Maybe `The official way to install Epinio` or something like that?

jimmykarily (approved):

PR #113, opened 2022-03-02
Add questions.yml to epinio single chart

juadk:
Fix https://github.com/epinio/epinio/issues/1247

This is a first glance, it could be improved.
When I was doing my tests, I've found an issue with external registry. I will track it in another issue.

It's not possible to add more conditional logic in the file, it's very limited (https://www.rancher.co.jp/docs/rancher/v2.x/en/catalog/custom/creating/#question-variable-reference)
So I tried to do it as clear as possible but we will likely adapt it after getting feedback.

**Note:** Cypress tests have to be updated

Installation only works in epinio namespace with epinio as deployment name (likely due to some hard-coded things):
image

### General settings screen:
image

### Advanced settings screen:
image

### External registry screen 1:
image
`Install local container registry` has to be unchecked to make external registry options appear (check next screen below)

### External registry screen 2:
image

### External S3 storage screen 1:
image
`Install minio` has to be unchecked to make external s3 storage options appear (check next screen below)

### External S3 storage screen 2:
image

### Debugging screen:
image

### DIFF
--- a/chart/epinio/questions.yml
+++ b/chart/epinio/questions.yml
@@ -1,0 +1,154 @@
+questions:
+- variable: email
+  label: Email
+  description: "Email to use for getting notifications about your certificates"
+  type: strings
+  required: false
+  group: "General settings"
+- variable: certManagerNamespace
+  label: Cert-manager namespace
+  description: "Namespace where cert-manager is installed in"
+  type: strings
+  required: false
+  group: "Advanced settings"
+- variable: ingress.ingressClassName
+  label: Ingress class name for the server
+  description: "Set a class name to select the ingress controller you want to use for the server"
+  type: strings
+  group: "Advanced settings"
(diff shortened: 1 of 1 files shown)


chart/epinio/questions.yml
```diff
@@ -0,0 +1,166 @@
...
+  type: boolean
+  show_subquestion_if: false
+  group: "External registry"
+  subquestions:
+  - variable: registry.url
+    label: External registry url
+    description: "URL of your external registry"
+    type: strings
+    required: false
+  - variable: registry.username
```
jimmykarily, line 96: shouldn't this be `global.registryUsername` ? same for the others in this section

juadk: oh yes good point, it's ok on my dev env and I forgot to change them here...

PR #115, opened 2022-03-03
Update epinio ui chart with latest build and env vars

richard-cox:
- Bump the image to 0.0.4
- Add an expiry time for the session (though until polling is replaced this probably won't be hit)
- Add an Epinio Version (which will be shown in the ui and should match the targetted epinio instance)

### DIFF
--- a/chart/epinio-ui/values.yaml
+++ b/chart/epinio-ui/values.yaml
@@ -1,5 +1,5 @@
 epinioUI:
-  image: ghcr.io/epinio/epinio-ui:v0.0.1-4
+  image: ghcr.io/epinio/epinio-ui:v0.0.1-5
   imagePullPolicy: IfNotPresent
 
 ingress:
--- a/chart/epinio-ui/values.yaml
+++ b/chart/epinio-ui/values.yaml
@@ -1,5 +1,5 @@
 epinioUI:
-  image: ghcr.io/epinio/epinio-ui:v0.0.1-3
+  image: ghcr.io/epinio/epinio-ui:v0.0.1-4
   imagePullPolicy: IfNotPresent
 
 ingress:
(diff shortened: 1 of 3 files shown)


olblak:
@richard-cox the backend version can be automated using this PR https://github.com/epinio/helm-charts/pull/118

olblak (dismissed):

olblak:
Btw you need to bump the chart version

chart/epinio-ui/values.yaml
```diff
@@ -18,6 +18,9 @@ epinioApiUrl: http://epinio-server.epinio.svc.cluster.local
 # Skip checking for valid SSL cert when making requests to `EPINIO_API_URL`
 epinioApiSkipSSL: "true"
 
+# This is the version that is displayed in the ui and should match that of the epinio it's targetting
+epinioVersion: "0.5.0"
```
olblak, line 22: I am bit confused by this setting. 
Could we be in a situation where the UI installed with Epinio 0.4?

olblak:
> Add an Epinio Version (which will be shown in the ui and should match the targetted epinio instance)

This one seems fragile to me as we'll quickly get outdated.
I think it would be better to get that information directly from Epinio

richard-cox:
@olblak I need to tie together all the versions out there (there's one for the UI frontend, another for the UI backend and finally one for the chart). Those in theory could all be independent from the Epinio version. The version in the chart is shown in the UI and is a kind of statement that the UI is compatible with a certain version of epinio. In theory someone could deploy the chart separately against a newer Epinio that the UI isn't compatible with. 

When it's included in the all in one chart I'm assuming this will be set via the parent chart

olblak (approved):

PR #116, opened 2022-03-03
 Fix unwanted irrelevant error messages from the staging scripts 

andreas-kupries:
ref https://github.com/epinio/epinio/issues/1240

fix: test custom cert file existence before use
fix: squash curl error output
fix: typo

### DIFF
--- a/chart/epinio/templates/stage-scripts.yaml
+++ b/chart/epinio/templates/stage-scripts.yaml
@@ -15,8 +15,9 @@ data:
     # internal s3 the chart has no information.  Therefore we cannot
     # use helm templating to insert these.
     echo By _ _ __ ___ _____ $(whoami) $(pwd)
-    cat /etc/ssl/certs/ca-bundle.crt > /tmp/ca-bundle.pem
-    cat /certs/ca.crt /certs/tls.crt >> /tmp/ca-bundle.pem
+    cat /etc/ssl/certs/ca-bundle.crt               > /tmp/ca-bundle.pem
+    test -f /certs/ca.crt   && cat /certs/ca.crt  >> /tmp/ca-bundle.pem
+    test -f /certs/tls.crt  && cat /certs/tls.crt >> /tmp/ca-bundle.pem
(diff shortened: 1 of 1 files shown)


chart/epinio/templates/stage-scripts.yaml
```diff
@@ -35,7 +36,7 @@ data:
     # command is not available in the `bash` image. The code as written now relies on each
     # unpacker to recognize/reject input properly.
     #
-    # The `sed` command add format indicators to the output from the unarchivers, making
+    # The `sed` commands add format indicators to the output from the unarchivers, making
```
olblak, line 39:
```suggestion
    # The `sed` command adds format indicators to the output from the unarchivers, making
```

chart/epinio/templates/stage-scripts.yaml
```diff
@@ -69,7 +70,7 @@ data:
     # is placed into a trap to ensure that it will always run, even for a staging failure.
     #
     set -e
-    trap "curl -X POST http://localhost:4191/shutdown || true" EXIT
+    trap "curl -X POST http://localhost:4191/shutdown 2> /dev/null || true" EXIT
```
olblak, line 73: What's the motivation to hid the curl command output?

chart/epinio/templates/stage-scripts.yaml
andreas-kupries: I put the plural on the command because there are several `sed`s, one per possible archive format.

chart/epinio/templates/stage-scripts.yaml
andreas-kupries: Linkerd may not be present (not installed, not up yet).
In that case the curl prints an error message and the user may be confused and believing that something is wrong.
The message is:

> curl: (7) Failed to connect to localhost port 4191: Connection refused

or similar.
As the situation is not truly an error the message is better suppressed.

andreas-kupries: :+1: sure.

andreas-kupries: Done, commit ede07bf70e394c3c45446f024f0432c131e70ada

juadk: I remember Loic and me wondering about this curl error message. Good thing to hide it.

juadk (approved):

PR #122, opened 2022-03-04
Switch slow workflows to self-hosted runners

manno:


jimmykarily (approved):

jimmykarily:
```
Run ct lint --config ct/ct.yaml
Linting charts...
Error: Error running process: exec: "helm": executable file not found in $PATH
Error running process: exec: "helm": executable file not found in $PATH
```

We need to install helm on the runners (or do it with some pipeline step)

manno:
We installed `unzip` on all self-hosted runners, retrying...

andreas-kupries (approved):

PR #125, opened 2022-03-07
Remove registry ingress

manno:
* removes registry ingress (`epinio-registry.$domain`)
* uses cluster DNS `registry.epinio-staging.svc.cluster.local` for internal registry
* enables TLS for internal registry
* adds sidecar for NodePort `127.0.0.1:30500`, so container runtime can access the registry via http (no need to trust our private CA)

fixes [#1276](https://github.com/epinio/epinio/issues/1276)
fixes https://github.com/epinio/helm-charts/issues/133

### DIFF
--- a/chart/epinio/templates/_helpers.tpl
+++ b/chart/epinio/templates/_helpers.tpl
@@ -1,5 +1,19 @@
+{{/*
+URL prefix for container images to be compatible with Rancher
+*/}}
 {{- define "registry-url" -}}
 {{- if .Values.registryURL -}}
 {{ trimSuffix "/" .Values.registryURL }}/
+{{- end -}}
+{{- end -}}
+
+{{/*
+URL of the registry epinio uses to store workload images
+*/}}
+{{- define "epinio.registry-url" -}}
+{{- if .Values.containerregistry.enabled -}}
+{{-   print "registry.epinio-staging.svc.cluster.local:5000" }}
+{{- else -}}
+{{-   print .Values.global.registryURL }}
 {{- end -}}
 {{- end -}}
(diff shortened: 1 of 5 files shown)


andreas-kupries:
This does not seem to have an epinio PR to test with, right ?
Would current epinio main be suitable ?

manno:
> This does not seem to have an epinio PR to test with, right ? Would current epinio main be suitable ?

Yes, I tested with Epinio main.
I started a PR to remove the `cahash` pkg from Epinio, but it's hard than I imagined, as the registry hostname contains a colon and that is not allowed in the staging job's  mount path.

manno:
The Kubernetes runtime in Rancher Desktop was able to connect to the TLS-enabled registry. It seems to skip TLS verification.

The runtime in k3d needed special configuration, because it [detects `localhost` in code](https://github.com/adisky/containerd/blob/main/pkg/cri/server/image_pull.go#L412-L418), switches to http and [enables insecureVerify](https://github.com/adisky/containerd/blob/main/pkg/cri/server/image_pull.go#L376-L381): 

```
mirrors:
  "127.0.0.1:30500":
    endpoint:
      - https://127.0.0.1:30500
configs:
  "127.0.0.1:30500":
    tls:
      insecure_skip_verify: true
```

Apparently none of these runtime benefited from the 'DaemonSet' workaround, to put the CA in `/etc/docker/cert.d`.

Instead of dealing with the peculiarities of container runtimes, we try to add a proxy to the registry, so:
* lifecycle/create can push via TLS
* container runtime can pull via http

andreas-kupries (approved):

PR #140, opened 2022-03-16
Fix some missing chart annotations (keywords, logo, sources, home)

andreas-kupries:
Fix https://github.com/epinio/epinio/issues/1300

Fixed:
  - missing license information (all)
  - missing keywords, source ref (ui, application)
  - missing logo, home (ui)
  - bogus image reference `curl` in the app chart's default `values.yaml`.

The last one should fix artifact hub notification noise for the application chart.
The message in the mails is
> error scanning image curl: image not found (package epinio-application:0.1.3)

### DIFF
--- a/chart/application/Chart.yaml
+++ b/chart/application/Chart.yaml
@@ -13,3 +13,5 @@ maintainers:
   name: SUSE
 sources:
 - https://github.com/epinio/helm-charts
+annotations:
+  artifacthub.io/license: Apache-2.0
--- a/chart/application/Chart.yaml
+++ b/chart/application/Chart.yaml
@@ -5,6 +5,11 @@ icon: https://raw.githubusercontent.com/epinio/helm-charts/main/assets/epinio.pn
 home: https://github.com/epinio/epinio
 type: application
 version: 0.1.13
+keywords:
+- epinio
+- paas
 maintainers:
 - email: team@epinio.io
   name: SUSE
+sources:
+- https://github.com/epinio/helm-charts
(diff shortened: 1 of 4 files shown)


andreas-kupries:
Ok. Learned a bit more about this from AH, now that I got the credentials to log in and see how we set things up there.

AH is not scanning the gh repo. It is scanning the helm repo we are publishing at https://epinio.github.io/helm-charts/ via gh-pages.

This is likely generated via the chart release workflow, triggered manually.

I will have to investigate how the tooling for that works to see if (and if yes, how) it is possible to retract a version.

The PR should fix the issue for the application chart going forward.
It is only the old versions which are at issue.

andreas-kupries:
Good. The proposed manual editing of the index.yaml worked. Because the chart releaser apparently only adds new things, and does not attempt to restore old things removed from the index. IOW cr looks only at the chart versions in the head of the branch and creates new releases if the branch is ahead of the index. It does not seem to look at old tags or releases to restore anything removed.

andreas-kupries:
And the repository now "contains" only `epinio-application` 0.1.14 (generated from this PR) as the sole release to be seen by AH.
It should not be noisy.
Also removed all the `epinio-installer` entries.

Ready for review.

juadk (approved):
I added the helm repo into my lab and I confirm I can not see epinio-installer anymore:
`helm repo add epinio https://epinio.github.io/helm-charts`
```
vagrant@k3s-opensusetw:~> helm search repo
NAME                     	CHART VERSION	APP VERSION	DESCRIPTION
epinio/container-registry	0.1.0        	           	A Helm chart for the Container Registry
epinio/epinio            	0.7.1        	v0.6.1     	The official way to install Epinio
epinio/epinio-application	0.1.13       	           	The helm chart for epinio applications to be de...
epinio/epinio-ui         	0.1.0        	           	A Helm chart for the Epinio UI
```
And looks good about the AH things.

PR #143, opened 2022-03-21
Make epinio-ui chart a subchart of epinio

### LINKED ISSUE #110: Make epinio-ui chart a subchart of epinio
Follow up of: https://github.com/epinio/ui/issues/68

manno:
fixes #110 

* disable ui ingress
* add explicit ui paths to epinio ingress


chart/epinio/templates/ingress.yaml
```diff
@@ -16,13 +16,44 @@ spec:
...
+        path: /wapi
+        pathType: ImplementationSpecific
+      - backend:
+          service:
+            name: epinio-server
+            port:
+              number: 80
+        path: /ready
+        pathType: ImplementationSpecific
+  {{- else }}
```
jimmykarily, line 48: Any reason why we can't limit Epinio endpoints to `/api` and `/wapi` always? In that case, we would simply add the `/` endpoint when `epinio-ui` is enabled.

jimmykarily (dismissed):

jimmykarily (approved):

PR #146, opened 2022-03-22
Run epinio server as non-root

manno:


jimmykarily (approved):

chart/epinio/templates/server.yaml
```diff
@@ -292,20 +292,20 @@ spec:
...
             httpGet:
               path: /ready
-              port: 80
-      # securityContext:
-      #   runAsNonRoot: true
-      #   runAsUser: 65532
+              port: 8030
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
```
jimmykarily, line 308: Does this UID need to match an existing user in the base image? Also I'm reading here: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod

```
 If this field is omitted, the primary group ID of the containers will be root(0).
```
Maybe we need to set `runAsGroup` too?

PR #148, opened 2022-03-23
Provide Image References for Staging Support To Kube Resources

andreas-kupries:
Ref https://github.com/epinio/epinio/issues/1305

Provide paketo/awscli/bash image reference, and supply to config map `epinio-stage-scripts`.

See https://github.com/epinio/epinio/pull/1311 for the PR changing the server to consume this information.

### DIFF
--- a/chart/epinio/templates/stage-scripts.yaml
+++ b/chart/epinio/templates/stage-scripts.yaml
@@ -4,6 +4,9 @@ metadata:
   name: epinio-stage-scripts
   namespace: {{ .Release.Namespace }}
 data:
+  builderImage:  "{{ template "registry-url" . }}{{ .Values.image.builder.repository}}:{{ .Values.image.builder.tag }}"
+  downloadImage: "{{ template "registry-url" . }}{{ .Values.image.awscli.repository}}:{{ .Values.image.awscli.tag }}"
+  unpackImage:   "{{ template "registry-url" . }}{{ .Values.image.bash.repository}}:{{ .Values.image.bash.tag }}"
   download: |-
     # Parameters
     # - PROTOCOL # s3 protocol
--- a/chart/epinio/templates/container-registry.yaml
+++ b/chart/epinio/templates/container-registry.yaml
@@ -106,7 +106,7 @@ spec:
       containers:
 {{ if .Values.containerregistry.createNodePort }}
       - name: nginx
-        image: nginx:1.21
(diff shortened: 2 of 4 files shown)


chart/epinio/templates/stage-scripts.yaml
```diff
@@ -4,6 +4,9 @@ metadata:
   name: epinio-stage-scripts
   namespace: {{ .Release.Namespace }}
 data:
+  builderImage:  "{{ template "registry-url" . }}{{ .Values.image.builder.repository}}:{{ .Values.image.builder.tag }}"
+  downloadImage: "{{ template "registry-url" . }}{{ .Values.image.awscli.repository}}:{{ .Values.image.awscli.tag }}"
+  unpackImage:   "{{ template "registry-url" . }}{{ .Values.image.bash.repository}}:{{ .Values.image.bash.tag }}"
```
jimmykarily, line 9: What about this? https://github.com/epinio/helm-charts/blob/main/chart/epinio/templates/container-registry.yaml#L109 `image: nginx:1.21`

jimmykarily:
`git grep "image:"` reveals a couple more images that don't respect the registry-url setting. It this PR only meant to fix the staging ones? To complete https://github.com/epinio/epinio/issues/1305 we will need to handle all images that prevent air-gapped installations.

chart/epinio/templates/stage-scripts.yaml
andreas-kupries: Thanks for the note. Will check.

> Is this PR only meant to fix the staging ones?

No. It should handle everything. The staging ones were simply the obvious ones seen when writing the ticket.

> `Values.images.epinio.repository` has the registry URL in it: `ghcr.io/epinio/epinio-server`.

Lovely. Not.

> It can still work if we agree that the repository values should not have a registry url in them when registry-url it set.
If that's the case, I think it deserves some comment in the values.yaml explaining that.

Is your proposal here to document the rule and then have the user adhere to it, manually ?

If yes, then that will __not__ work with Rancher Marketplace AFAIK, because RM will simply set the registry url (Actually `.Values.global.cattle.systemDefaultRegistry`), and certainly not manipulate the repository value at all.

I suspect that this requires me to pull the `ghcr.io` part out of `repository` into a separate `registry` value, and then rework the template to auto-choose from the registry url vs that, with the former having preference over the latter.

andreas-kupries: Pushed fixes, see 3865845

jimmykarily (approved):

PR #153, opened 2022-03-31
Fixed and extended epinio values schema

andreas-kupries:
- Fixed missing entry for new field `image.epinio.registry`.
  - Added schema for the subchart configuration hierarchies.

:notes: Re-indented using emacs javascript mode.

### DIFF
--- a/chart/epinio/values.schema.json
+++ b/chart/epinio/values.schema.json
@@ -180,9 +180,6 @@
     "containerregistry": {
       "type": "object",
       "properties": {
-        "createNodePort": {
-          "type": "boolean"
-        },
         "enabled": {
           "type": "boolean"
         },
@@ -225,7 +222,6 @@
         }
       },
       "required": [
-        "createNodePort",
         "enabled",
         "image",
         "imagePullPolicy",
--- a/chart/epinio/values.schema.json
+++ b/chart/epinio/values.schema.json
@@ -181,5 +181,205 @@
   "required": [
     "certManagerNamespace",
     "s3"
-  ]
+  ],
+  "containerregistry": {
+    "type": "object",
+    "properties": {
+      "createNodePort": {
+        "type": "boolean"
+      },
+      "enabled": {
+        "type": "boolean"
+      },
+      "image": {
+        "type": "object",
+        "properties": {
+          "nginx": {
+            "type": "object",
+            "properties": {
+              "repository": {
+                "type": "string"
+              },
+              "tag": {
+                "type": "string"
+              }
+            }
+          },
+          "registry": {
+            "type": "object",
+            "properties": {
+              "repository": {
+                "type": "string"
+              },
+              "tag": {
+                "type": "string"
+              }
+            }
+          }
+        },
+        "required": [
+          "nginx",
+          "registry"
+        ]
+      },
+      "imagePullPolicy": {
+        "type": "string"
+      },
+      "ingressClassName": {
+        "type": "string"
+      }
+    },
+    "required": [
+      "createNodePort",
+      "enabled",
+      "image",
+      "imagePullPolicy",
+      "ingressClassName"
+    ]
+  },
+  "epinio-ui": {
+    "type": "object",
+    "properties": {
+      "enabled": {
+        "type": "boolean"
+      },
+      "ingress": {
+        "type": "object",
+        "properties": {
+          "enabled": {
+            "type": "boolean"
+          }
+        },
+        "required": [
+          "enabled"
+        ]
+      }
+    },
+    "required": [
+      "enabled",
+      "ingress"
+    ]
+  },
+  "kubed": {
+    "type": "object",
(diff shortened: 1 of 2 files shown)


chart/epinio/values.schema.json
```diff
@@ -1,177 +1,380 @@
...
+    "properties": {
+	"registryURL": {
+	    "type": "string"
+	},
+	"image": {
+	    "type": "object",
+	    "properties": {
+		"epinio": {
+		    "type": "object",
+		    "properties": {
```
olblak, line 14: Nothing critical but indentation is broken  :)

olblak (dismissed):
Excepted the indentation, it looks good to me

chart/epinio/values.schema.json
andreas-kupries: You saw commit https://github.com/epinio/helm-charts/pull/153/commits/7f67eb1f21712ac99af9cf8251a36ec5297fc1fa ?
What indentation mode was used before that ?

manno:
We understood from the documentation that validations should be in the sub-charts preferably. However our sub-charts don't have any, so it's good to add it here.

Also this adds tabs (8 whitespaces) for indentation. I'd prefer to keep using spaces.

jimmykarily:
@andreas-kupries we found out with @manno  that the values schema was wrong (fixed here: https://github.com/epinio/helm-charts/commit/f51b81aec8a5009c23b5b428b64b212d7efdd370). Also, [according to docs](https://helm.sh/docs/topics/charts/#schema-files): 

```
Furthermore, the final .Values object is checked against all subchart schemas. This means that restrictions on a subchart can't be circumvented by a parent chart. This also works backwards - if a subchart has a requirement that is not met in the subchart's values.yaml file, the parent chart must satisfy those restrictions in order to be valid.
```

This makes me think  that these validations belong to the subchart schemas. We don't control those, so maybe we can skip these?

chart/epinio/values.schema.json
andreas-kupries: Ah, that is an issue with `diff` output (`[+-]`) weirdly interacting with tabs. (The effect helps when looking at YAML, unwanted tabs become more obvious)

andreas-kupries:
After the rebase and conflict resolution the only true change left is the schema work for the sub charts. Given the comment

> This makes me think that these validations belong to the subchart schemas. But we can keep them since we don't know what the subcharts do.

I am now more trending towards that this PR (and ticket) should be abandoned. I.e. if it had anything left beyond the subchart schema work (SCSW) we could argue for keeping the SCSW. with SCSW the only think it may be better to simply not add it all.

Thoughts ? @jimmykarily @manno

manno:
Added a commit, which moves some nodes below the first property “Values”


I have this broken values.yaml:

```
USER-SUPPLIED VALUES:

kubed:
  enabled: no
  fullnameOverride: true

epinio-ui:
  enabled:
    - no

global:
  domain:
    - localdev.me
  cattle:
    systemDefaultRegistry: 123

ingress:
  ingressClassName: nginx
  nginxSSLRedirect: false
```
I do 
`helm lint --with-subcharts --strict --values values.yml chart/epinio` (helm v3.8.2) and I see these errors:

```
dependencies.go:49: Warning: Condition path 'epinio-ui.enabled' for chart epinio-ui returned non-bool value
==> Linting chart/epinio
[ERROR] values.yaml: - global.cattle.systemDefaultRegistry: Invalid type. Expected: string, given: integer
- global.domain: Invalid type. Expected: string, given: array
- epinio-ui.enabled: Invalid type. Expected: boolean, given: array
- kubed.fullnameOverride: Invalid type. Expected: string, given: boolean
- minio.makeUserJob.podAnnotations.linkerd.io/inject: Invalid type. Expected: object, given: string

[ERROR] templates/: values don't meet the specifications of the schema(s) in the following chart(s):
epinio:
- minio.makeUserJob.podAnnotations.linkerd.io/inject: Invalid type. Expected: object, given: string
- global.cattle.systemDefaultRegistry: Invalid type. Expected: string, given: integer
- global.domain: Invalid type. Expected: string, given: array
- epinio-ui.enabled: Invalid type. Expected: boolean, given: array
- kubed.fullnameOverride: Invalid type. Expected: string, given: boolean


==> Linting chart/epinio/charts/epinio-ui-0.1.2.tgz
[ERROR] templates/: template: epinio-ui/templates/_helpers.tpl:70:25: executing "epinio-ui.registry" at <.Values.global.cattle.systemDefaultRegistry>: wrong type for value; expected string; got float64

==> Linting chart/epinio/charts/kubed-v0.13.2.tgz

==> Linting chart/epinio/charts/minio-3.6.3.tgz

Error: 4 chart(s) linted, 2 chart(s) failed
```

The remaining problem is the missing default value for `containerregistry.createNodePort`

jimmykarily (approved):

PR #173, opened 2022-04-15
Add rabbitmq dev catalog service

jimmykarily:


chart/epinio/templates/service-catalog.yaml
```diff
@@ -57,4 +57,22 @@ spec:
...
+---
+apiVersion: application.epinio.io/v1
+kind: Service
+metadata:
+  name: rabbitmq-dev
+  namespace: {{ .Release.Namespace }}
+spec:
+  name: rabbitmq-dev
+  shortDescription: A RabbitMQ service that can be used during development
+  description: "This service is going to deploy a simple default Bitnami RabbitMQ db
```
enrichman, line 69: I guess that the `db` is a leftover. Also the `\n`.

enrichman (approved):

PR #176, opened 2022-04-19
Update NOTES

enrichman:
Updated NOTES removing instructions on how to get the local ingress IP (this is system and cluster dependent).

Fix [#134](https://github.com/epinio/epinio/issues/1343)

### DIFF
--- a/chart/epinio/templates/NOTES.txt
+++ b/chart/epinio/templates/NOTES.txt
@@ -1,15 +1,7 @@
-
-{{- if .Values.global.domain }}
-You can get the IP address of your Ingress controller by passing this command:
-`kubectl describe svc traefik --namespace traefik | grep Ingress | awk '{print $3}'`
-
-Make sure your domain points to the IP address of your Ingress controller (1.2.3.4).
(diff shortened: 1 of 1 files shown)


chart/epinio/templates/NOTES.txt
```diff
@@ -1,15 +1,7 @@
-
-{{- if .Values.global.domain }}
-You can get the IP address of your Ingress controller by passing this command:
-`kubectl describe svc traefik --namespace traefik | grep Ingress | awk '{print $3}'`
-
-Make sure your domain points to the IP address of your Ingress controller (1.2.3.4).
```
juadk, line 6: Do we really want to remove this message? because the user has to think about updating its DNS with the correct IP. 
When we talked about it this morning, I though we wanted to remove the `kubectl` command and just say something like `Get the IP address of your Ingress controller and update your DNS accordingly.`

juadk (approved):
Only one comment, otherwise LGTM.

juadk (approved):

PR #188, opened 2022-05-04
updating documentation

mmartin24:
Updating dead links for Epinio installation on helm-charts


mmartin24:
> I think it could be better pointing to the main page `https://docs.epinio.io`/ instead of `https://docs.epinio.io/installation` because we don't mention installation specifically but the global documentation.

perfect.
Just did so.

PR #190, opened 2022-05-09
Remove leading digits from resource names

jimmykarily:
Fixing the problem that was introduced here: https://github.com/epinio/helm-charts/pull/187/files

We used to add prefixes like `i-` for Ingress and such so we didn't get leading digits. We will now just remove the leading digits.

### DIFF
--- a/chart/application/templates/_helpers.tpl
+++ b/chart/application/templates/_helpers.tpl
@@ -24,9 +24,8 @@ app.kubernetes.io/name: {{ .Values.epinio.appName }}
 app.kubernetes.io/component: application
 {{- end }}
 
-
 {{/*
-Remove character that are invalid for kubernetes resource names from the
+Removes character that are invalid for kubernetes resource names from the
 given string
 */}}
 {{- define "epinio-name-sanitize" -}}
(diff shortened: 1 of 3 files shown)


jimmykarily:
Hm I just realized that a digit-only name would result in a resource name starting with a `-` (since all characters would be removed). This would then be invalid :facepalm: .

jimmykarily:
Opted for adding an `r` prefix to ensure the name doesn't start with a digit.

chart/application/templates/_helpers.tpl
```diff
@@ -24,9 +24,8 @@ app.kubernetes.io/name: {{ .Values.epinio.appName }}
 app.kubernetes.io/component: application
 {{- end }}
 
-
 {{/*
-Remove character that are invalid for kubernetes resource names from the
+Removes character that are invalid for kubernetes resource names from the
```
andreas-kupries, line 28:
```suggestion
Removes characters that are invalid for kubernetes resource names from the
```

chart/application/templates/_helpers.tpl
```diff
@@ -35,10 +34,11 @@ given string
 
 {{/*
 Resource name sanitization and truncation.
-Always suffix the sha1sum (40 characters long)
-The rest of the character up to 63 is the original string with invalid
+- Always suffix the sha1sum (40 characters long)
+- Always add an "r" prefix to make sure we don't have leading digits
+- The rest of the character up to 63 is the original string with invalid
```
andreas-kupries, line 39:
```suggestion
- The rest of the characters up to 63 are the original string with invalid
```

andreas-kupries (changes requested):

andreas-kupries (approved):

PR #199, opened 2022-05-12
Helm Controller and CRDs

enrichman:
#### Issue
- https://github.com/epinio/epinio/issues/1398
---

This PR adds the Helm Controller in clusters without it already installed.

The `helm-controller` will be installed if the two CRDs are missing.
The HelmChart and HelmChartConfig were generated from the [k3s-io/helm-controller](https://github.com/k3s-io/helm-controller) repo, with `controller-gen` (the original CRDs in the repo are too old):

```
controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
controller-gen "crd:trivialVersions=true,preserveUnknownFields=false" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```

The `helm-controller` will be deployed in the `kube-system` namespace. I've tried to deploy it in a different one (such as `epinio`) and it worked, but I had to give to it very large permissions (probably the same that it has in the `kube-system`).

I got this error while trying to tune the minimum set of permissions:

```
user "system:serviceaccount:epinio:helm-controller" (groups=["system:serviceaccounts" "system:serviceaccounts:epinio"  "system:authenticated"])
is attempting to grant RBAC permissions not currently held:
{
  APIGroups: ["*"],
  Resources:["*"],
  Verbs:["*"]
}
{
  NonResourceURLs: ["*"],
  Verbs: ["*"]
}  
```

so it needs to do everything everywhere. This is because Helm doesn't know in advance what kind of resources it needs to create.

Anyway, if we want to deploy it in the `epinio` namespace we just need to set the namespace and the service account in the Deployment:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: helm-controller
spec:
  template:
    spec:
      serviceAccountName: helm-controller
```
and then the ClusterRole, ServiceAccount and ClusterRoleBinding:
```yaml
# RBAC - ClusterRoleBinding
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: helm-controller
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: helm-controller
subjects:
- kind: ServiceAccount
  name: helm-controller

# RBAC - ServiceAccount
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: helm-controller

# RBAC - ClusterRole auth
---
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: helm-controller
rules:
- apiGroups: ["*"]
  resources: ["*"]
  verbs: ["*"]
```

### DIFF
--- a/chart/epinio/templates/helm-controller.yaml
+++ b/chart/epinio/templates/helm-controller.yaml
@@ -1,7 +1,4 @@
-{{- if and
-      (not (.Capabilities.APIVersions.Has "helmcharts.helm.cattle.io/v1"))
-      (not (.Capabilities.APIVersions.Has "helmchartconfigs.helm.cattle.io/v1"))
--}}
+{{- if not (.Capabilities.APIVersions.Has "helm.cattle.io/v1") -}}
 
 # Helm Controller Deployment
 ---
--- a/chart/epinio/templates/helm-controller.yaml
+++ b/chart/epinio/templates/helm-controller.yaml
@@ -24,44 +24,6 @@ spec:
           image: rancher/helm-controller:v0.12.1
           command: ["helm-controller"]
 
-# RBAC - ClusterRoleBinding
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
-metadata:
-  name: helm-controller
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: helm-controller
-subjects:
-- kind: ServiceAccount
-  name: helm-controller
-  namespace: {{ .Release.Namespace }}
-
-# RBAC - ServiceAccount
----
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: helm-controller
-  namespace: {{ .Release.Namespace }}
-
-# RBAC - ClusterRole auth
----
-kind: ClusterRole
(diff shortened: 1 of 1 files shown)


chart/epinio/templates/helm-controller.yaml
```diff
@@ -0,0 +1,155 @@
+{{- if not (.Capabilities.APIVersions.Has "helm.cattle.io/v1") -}}
```
jimmykarily, line 1: This Deployment doesn't specify a serviceAccountName which means it runs as the default serviceaccount in the kube-system namespace. This serviceaccount may or may not have cluster-admin role bound to it: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#service-account-permissions

If we need to be cluster-admins, we have to be explicit about it. To avoid messing around with the default serviceaccount, we should create a serviceaccount and bind the cluster-admin role to it. Then explicitly use it here. I'll give that a try.

jimmykarily: Done in c2c4ed634ee9e3d8d33fbf7bd2cfeb89ef67f800

andreas-kupries (approved):

PR #200, opened 2022-05-12
Add helm hooks to cleanup catalog service PVCs

jimmykarily:
Fixes https://github.com/epinio/epinio/issues/1386

Accompanying PR: https://github.com/epinio/epinio/pull/1441


andreas-kupries (approved):
Ok, so the bitnami charts have a hook in the form of the `extraDeploy` key where the operator can specify additional resources to be deployed with the base chart. And we are using this to add a hook job which removes the PVCs the chart has created when it was deployed.

That is nice. (And nice from bitnami to actually provide such an extension mechanism)

PR #208, opened 2022-05-30
Use Bcrypt for passwords

enrichman:
Ref:
- https://github.com/epinio/epinio/issues/1361
---

This PR will store the passwords as Bcrypt hashes instead of plaintext.

### DIFF
--- a/chart/epinio/templates/NOTES.txt
+++ b/chart/epinio/templates/NOTES.txt
@@ -1,8 +1,8 @@
 To interract with your Epinio installation download the latest epinio binary from https://github.com/epinio/epinio/releases/latest.
 
-Update the api location and credentials with:
+Login to the cluster with:
     
-    `epinio settings update`
+    `epinio login -u admin -p password https://epinio.{{ .Values.global.domain }}`
 
 or go to the dashboard at: https://epinio.{{ .Values.global.domain }}
 
--- a/chart/epinio/templates/NOTES.txt
+++ b/chart/epinio/templates/NOTES.txt
@@ -2,8 +2,10 @@ To interract with your Epinio installation download the latest epinio binary fro
 
 Login to the cluster with:
     
-    `epinio login -u admin -p password https://epinio.{{ .Values.global.domain }}`
+    `epinio login -u {{.Values.api.username}} https://epinio.{{ .Values.global.domain }}`
 
 or go to the dashboard at: https://epinio.{{ .Values.global.domain }}
+
+If you didn't specify a password the default one is `password`.
 
 For more information about Epinio, feel free to checkout https://epinio.io/ and https://docs.epinio.io/.
--- a/chart/epinio/templates/default-user.yaml
+++ b/chart/epinio/templates/default-user.yaml
@@ -22,6 +22,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 stringData:
   username: epinio
+  # bcrypt hash for 'password'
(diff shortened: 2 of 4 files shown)


chart/epinio/templates/NOTES.txt
```diff
@@ -1,8 +1,8 @@
 To interract with your Epinio installation download the latest epinio binary from https://github.com/epinio/epinio/releases/latest.
 
-Update the api location and credentials with:
+Login to the cluster with:
     
-    `epinio settings update`
+    `epinio login -u admin -p password https://epinio.{{ .Values.global.domain }}`
```
jimmykarily, line 5: Unless the user set these values: https://github.com/epinio/helm-charts/blob/5d947c0b9c26c2b2254ee311d0d4d96ced0291c7/chart/epinio/values.yaml#L60-L64

Maybe we should say something like:

```
`epinio login -u {{.Values.api.username}} -p api_password_here https://epinio.{{ .Values.global.domain }}`

If you didn't specify a password the default one is `password`.
```

chart/epinio/values.yaml
```diff
@@ -60,8 +60,8 @@ s3:
 api:
   # Default user username
   username: admin
-  # Default user password
-  password: password
+  # Default user password (bcrypt hash for 'password')
+  password: "$2a$10$6bCi5NMstMK781In7JGiL.B44pgoplUb330FQvm6mVXMppbXBPiXS"
```
jimmykarily, line 64: I feel the name of this value should change to something like `encryptedPassword` or even `passwordBcrypt`. Another option would be to let the password be plain text in values and we do the encryption in helm (if supported). But in that case, the plain-text password will float around in values.yaml files and command line histories, so let's not.

chart/epinio/templates/default-user.yaml
```diff
@@ -22,6 +22,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 stringData:
   username: epinio
-  password: password
+  # bcrypt hash for 'password'
+  password: "$2a$10$6bCi5NMstMK781In7JGiL.B44pgoplUb330FQvm6mVXMppbXBPiXS"
```
jimmykarily, line 26: Not relevant to this PR but I just realized we always create this default user with the dummy `password`. Operators have no option to skip it, which would make perfect sense in production environments, where they would create users through other means. I created an issue here: https://github.com/epinio/helm-charts/issues/214

chart/epinio/templates/NOTES.txt
enrichman: What about not having the password at all as a flag? In that case the user will be prompted for it
```
`epinio login -u {{.Values.api.username}} https://epinio.{{ .Values.global.domain }}`

If you didn't specify a password the default one is `password`.
```
 i.e.:
```
-> % epinio login -u admin https://epinio.172.19.0.4.omg.howdoi.website

🚢  Login to your Epinio cluster [https://epinio.172.19.0.4.omg.howdoi.website]
Password: 

⚠️  Certificate signed by unknown authority

|     KEY     |       VALUE       |
|-------------|-------------------|
| Issuer Name | CN=epinio-ca      |
| Common Name | epinio-ca         |
| Expiry      | 2022-September-01 |

Do you want to trust it (y/n): y

✔️  Trusting certificate...

✔️  Login successful
```

jimmykarily: yes, that would do.

chart/epinio/values.yaml
enrichman: Renamed to `passwordBcrypt` just to make it more explicit the algorithm used

jimmykarily (approved):

PR #213, opened 2022-06-03
Image references, Styling, Simplification, Fixes

andreas-kupries:
Ref https://github.com/epinio/epinio/issues/1520
Ref https://github.com/epinio/epinio/issues/1512

Note: The simplification of the app chart is not truly tested yet. Needs a change to the chart ref and locally made chart tarball.

### DIFF
--- a/chart/epinio-ui/templates/server.yaml
+++ b/chart/epinio-ui/templates/server.yaml
@@ -40,9 +40,9 @@ spec:
         - name: ALLOWED_ORIGINS
           value: {{ default (printf "https://epinio.%s" .Values.global.domain) .Values.epinioAllowedOrigins }}
         - name: EPINIO_API_URL
-          value: {{ default (printf "http://epinio-server.%s.svc.cluster.local" .Release.Namespace) .Values.epinioAPIUrl }}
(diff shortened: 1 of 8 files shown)


chart/epinio-ui/templates/server.yaml
```diff
@@ -38,11 +38,11 @@ spec:
 
         env:
         - name: EPINIO_API_URL
-          value: {{ default (printf "http://epinio-server.%s.svc.cluster.local" .Release.Namespace) .Values.epinioApiUrl }}
+          value: {{ default (printf "http://epinio-server.%s.svc.cluster.local" .Release.Namespace) .Values.epinioAPIUrl }}
```
jimmykarily, line 41: Are we following [golang conventions](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)? In that case, it should be `epinioAPIURL` or `epinioAPIurl` (I prefer the first).

andreas-kupries: Sure, lets go full golang

andreas-kupries: Done. Commit 750341131b59f564ad6eba925a24508a5784d7b8

jimmykarily (approved):

PR #219, opened 2022-06-10
fix: revert bash image used in staging steps.

andreas-kupries:
Ref https://github.com/epinio/epinio/issues/1541

rancher/shell uses user `shell:1000`, whereas `library/bash` uses `root:0`.
This is important for calls of `chown`.


juadk (approved):
LGTM, tested successfully on GKE.

PR #220, opened 2022-06-14
Do not automate epinioVersion update

olblak:
It appears that epinioVersion is not used anymore which break updatecli execution as showned [here|https://github.com/epinio/helm-charts/actions/workflows/updatecli.yml]

@richard-cox Could you confirm that we can disble this automation


richard-cox (approved):
There is now a manual step to set this to the epinio chart version (https://github.com/epinio/epinio/pull/1548)

jimmykarily (approved):

PR #241, opened 2022-06-23
Add pipeline and files to create a docker image for each new release

jimmykarily:
Possibly fixes: https://github.com/epinio/epinio/issues/1589

### DIFF
--- a/.github/workflows/build-epiniod.yml
+++ b/.github/workflows/build-epiniod.yml
@@ -1,6 +1,7 @@
 name: Build epiniod image
 
 on:
+  workflow_dispatch:
   release:
     types: [published]
 jobs:


.github/workflows/build-epiniod.yml
```diff
@@ -0,0 +1,36 @@
+name: Build epiniod image
+
+on:
+  release:
+    types: [published]
```
jimmykarily, line 5: Maybe we can also add the workflow dispatch trigger? This will allow us to try out the pipeline manually.

andreas-kupries (dismissed):

.github/workflows/build-epiniod.yml
jimmykarily: added that

PR #243, opened 2022-06-27
Initialize upgrade workflow to test upgrade

juadk:
This PR adds a workflow to test upgrading epinio helmchart.
I don't think it's the final version but I need to initialize a new workflow if I want to test it targeting git branches.
It won't break anything because it only adds new files in the repo, no current process is touched.


juadk:
Finally, the test will be executed from epinio/epinio

PR #247, opened 2022-07-08
1619 bump dependencies

jimmykarily:
https://github.com/epinio/epinio/issues/1619

Consuming PR: https://github.com/epinio/epinio/pull/1628/files


chart/epinio/values.yaml
```diff
@@ -65,6 +65,8 @@ api:
 # Minio subchart values
 minio:
   enabled: true
+  image:
+    tag: RELEASE.2022-07-06T20-29-49Z
```
jimmykarily, line 69: This is overriding the default value from the helm chart: https://github.com/minio/minio/blob/4bfe849409a2bdb597d33a4a498ab78b1f8eb1df/helm/minio/values.yaml#L17

because that's where the warning comes from:

image

I'm afraid it's easy to forget to change this value the next time we bump the minio chart and we will keep using a subbed version of the image. Also, I'm not sure to what degree this image is compatible with the rest of the chart. Our tests are passing though ([see CI on this PR](https://github.com/epinio/epinio/issues/1619))

jimmykarily: I've opened a PR upstream: https://github.com/minio/minio/pull/15255

PR #175, opened 2022-04-18
Additional application charts

agracey:
For: https://github.com/epinio/epinio/issues/1273

Still TODO:
- [ ] Test HPA and document requirements
- [ ] Test failure state when HPA or Metrics Server are not installed in cluster
- [ ] Validate NetworkPolicy structure
- [ ] Document usage of both

PR #255, opened 2022-07-28
App standard chart for stateful apps

andreas-kupries:
Ref https://github.com/epinio/epinio/issues/1658

This is branched of from #175.
The other experimental app charts are removed here, so that only the new stateful app chart is released by the action.

### DIFF
--- a/chart/application-stateful/.helmignore
+++ b/chart/application-stateful/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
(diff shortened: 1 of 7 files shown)


andreas-kupries:
This needs changes in the generation of sts names, to ensure they are short enough so that adding a hash to them for the `controller-revision-hash` does not overflow 64 characters.

Unassigned and moved back to todo for others to take, should they desire, while I am on vacation.

andreas-kupries:
Out of review: Do not distribute the chart with epinio. I.e. not pre-installed. Use it only in the epinio tests to demo that epinio can work with an app not having a Deployment as main resource. For that adding it as a transient custom chart during the tests is enough.

jimmykarily (approved):

PR #259, opened 2022-08-11
Move `created-by` from label to annotation

enrichman:
Moving the user from the `app.kubernetes.io/created-by` label to the `epinio.io/created-by` annotation to avoid the 63 char limitations.

It also adds a user with an email.

### DIFF
--- a/chart/epinio/values.yaml
+++ b/chart/epinio/values.yaml
@@ -60,12 +60,6 @@ api:
       role: user
       workspaces:
         - workspace
-    - username: user1@epinio.io
-      passwordBcrypt: "$2a$10$6bCi5NMstMK781In7JGiL.B44pgoplUb330FQvm6mVXMppbXBPiXS"
-      role: user
-      workspaces:
-        - workspace
-
 # Minio subchart values
 minio:
   enabled: true
--- a/chart/epinio/values.yaml
+++ b/chart/epinio/values.yaml
@@ -60,6 +60,12 @@ api:
       role: user
       workspaces:
         - workspace
+    - username: user1@epinio.io
+      passwordBcrypt: "$2a$10$6bCi5NMstMK781In7JGiL.B44pgoplUb330FQvm6mVXMppbXBPiXS"
+      role: user
+      workspaces:
+        - workspace
+        
 # Minio subchart values
 minio:
   enabled: true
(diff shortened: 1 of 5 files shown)


chart/epinio/templates/_helpers.tpl
```diff
@@ -114,7 +114,7 @@ extraDeploy:
       labels:
         app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
         app.kubernetes.io/instance: {{ .Release.Name | quote }}
-        app.kubernetes.io/version: {{ default .Chart.AppVersion .Values.image.epinio.tag }}
```
jimmykarily, line 117: Should we remove this value then? https://github.com/epinio/helm-charts/blob/75b8fae6baf7734aa37012d8b1377e9d8a903313/chart/epinio/values.yaml#L11

I think we use it during development to point to the image built locally: https://github.com/epinio/epinio/blob/83b68f9510a3f1681eec0ef558a9f26188e14087/scripts/prepare-environment-k3d.sh#L84

chart/epinio/values.yaml
```diff
@@ -60,6 +60,12 @@ api:
       role: user
       workspaces:
         - workspace
+    - username: user1@epinio.io
+      passwordBcrypt: "$2a$10$6bCi5NMstMK781In7JGiL.B44pgoplUb330FQvm6mVXMppbXBPiXS"
+      role: user
+      workspaces:
+        - workspace
```
jimmykarily, line 67: Why do we need an additional user?

chart/epinio/templates/_helpers.tpl
enrichman: Ah, I guess i forgot to rebase! Missed it, thanks.

chart/epinio/values.yaml
enrichman: I added it to test a user with an email. If you think it's not needed I can remove it. :)

chart/epinio/templates/_helpers.tpl
enrichman: Ah no, this is correct. This one was the only place where we don't need it.

jimmykarily (approved):

chart/epinio/values.yaml
jimmykarily: It would end up in everybody's cluster by default. If we need it in tests, we should create it in tests.

PR #263, opened 2022-08-17
Use explicit pathType values for Ingress

jimmykarily:
to avoid unexpected implementations. Also move the "catch all" rule at
the end because in some cases it seems to matter (See:
https://github.com/epinio/epinio/issues/1663#issuecomment-1206834066)

Sibling PR: https://github.com/epinio/epinio/pull/1685

### DIFF
--- a/chart/epinio/templates/ingress.yaml
+++ b/chart/epinio/templates/ingress.yaml
@@ -21,6 +21,27 @@ spec:
   - host: "epinio.{{ .Values.global.domain }}"
     http:
       paths:
+      - backend:
+          service:
+            name: epinio-server
+            port:
+              number: 80
+        path: /api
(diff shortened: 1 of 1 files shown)


andreas-kupries:
@jimmykarily Should we rebase and update the `standard-stateful` chart also ? Or do that separately ?

andreas-kupries:
Ignore that comment please. I seem to have confused the system ingress changed here with the app ingresses from the app chart.

jimmykarily:
> Ignore that comment please. I seem to have confused the system ingress changed here with the app ingresses from the app chart.

either this, or I had actually changed the application chart too and later reverted, I don't remember. All good.

andreas-kupries (approved):
looks ok

PR #264, opened 2022-08-19
Support app(chart) customization

andreas-kupries:
chore: updated app(chart) crds

### DIFF
--- a/chart/epinio/crds/app-crd.yaml
+++ b/chart/epinio/crds/app-crd.yaml
@@ -82,6 +82,12 @@ spec:
                 items:
                   type: string
                 type: array
+              settings:
+                additionalProperties:
+                  type: string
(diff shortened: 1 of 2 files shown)


jimmykarily (approved):

jimmykarily:
Shouldn't this PR add `userConfig` in [the values.yaml](https://github.com/epinio/helm-charts/blob/main/chart/application/values.yaml)?

jimmykarily:
> Shouldn't this PR add `userConfig` in [the values.yaml](https://github.com/epinio/helm-charts/blob/main/chart/application/values.yaml)?

hm, since we make no use of that value in the chart, it's not needed. We can do it if we decide to merge the stateful and stateless application charts. Ignore it for now.

PR #266, opened 2022-08-26
1634 custom registry cert

jimmykarily:
Sibling PR https://github.com/epinio/epinio/pull/1699

### DIFF
--- a/chart/epinio/templates/container-registry.yaml
+++ b/chart/epinio/templates/container-registry.yaml
@@ -3,7 +3,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: auth
+  name: registry-auth
   namespace: {{ .Release.Namespace }}
 stringData:
   # The only supported password format is bcrypt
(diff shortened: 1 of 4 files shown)


jimmykarily:
Turns out, the changes in this PR are not enough. When the app is deployed we get this error on the Deployment object:

```
 image "registry.epinio.svc.cluster.local:5000/apps/workspace-sample:5fa235d2f2b36191": failed to resolve reference "registry.epinio.svc.cluster.local:5000/apps/workspace-sample:5fa235d2f2b36191": failed to do │
│  request: Head "https://registry.epinio.svc.cluster.local:5000/v2/apps/workspace-sample/manifests/5fa235d2f2b36191": dial tcp: lookup registry.epinio.svc.cluster.local: no such host 
```

The registry is reachable only from withing the cluster itself. Disabling the "localhost" hack, means the registry has to be exposed in some other way for the kubernetes to access it. One option is to keep the "localhost" hack but with SSL this time.

jimmykarily:
localhost is treated as "insecure" by default: https://github.com/adisky/containerd/blob/main/pkg/cri/server/image_pull.go#L376

not sure if there is a way to force TLS when the registry is on localhost

jimmykarily:
Turns out to be more complicated than it's worth it ([read my comment here](https://github.com/epinio/epinio/issues/1634#issuecomment-1230321459)). I'm closing this PR.

PR #267, opened 2022-09-01
Introduce extraEnv for server deployment

thehejik:
This PR brings support for setting an extra environment variable(s) in epinio-server deployment, also more than one extraEnv variable/value can be set by using an indexes.

Usage:
```
 helm upgrade ... --set "extraEnv[0].name=SESSION_KEY" --set-string "extraEnv[0].value=12345"
```

This is first part addressing https://github.com/epinio/epinio/issues/1675

### DIFF
--- a/chart/epinio/templates/server.yaml
+++ b/chart/epinio/templates/server.yaml
@@ -288,6 +288,9 @@ spec:
             - name: INGRESS_CLASS_NAME
               value: "{{ .Values.ingress.ingressClassName }}"
             {{- end }}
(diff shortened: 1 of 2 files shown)


thehejik:
I had to handle extra newlines due `nindent` - now it renders correctly without a newlines before name/value yaml block (even when the `extraEnv[0].name` is not used or has no value assigned).

andreas-kupries (dismissed):

andreas-kupries (approved):

PR #275, opened 2022-09-14
Change how we set the epiniod image tag

### LINKED ISSUE #250: Fix epiniod pipeline
After releasing v1.0.1 of the helm chart, [the epiniod pipeline]() didn't trigger automatically. I had to [trigger it manually](https://github.com/epinio/helm-charts/runs/7252226597?check_suite_focus=true) and it seemed to have released something but the tag is wrong:

```
docker run -it --privileged  ghcr.io/epinio/epiniod:epinio-1.0.1
...
Release "epinio" does not exist. Installing it now.
Error: failed to download "epinio/epinio" at version "epinio-1.0.1"
...
```

Not sure why, but this doesn't seem to work: https://github.com/epinio/helm-charts/blob/e64ad567140601d4ec515fa4b1f67e4f3abedb6d/.github/workflows/build-epiniod.yml#L5-L6

jimmykarily:
Fixes #250

### DIFF
--- a/.github/workflows/build-epiniod.yml
+++ b/.github/workflows/build-epiniod.yml
@@ -27,25 +27,21 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      # https://github.com/docker/build-push-action/blob/master/docs/advanced/tags-labels.md
-      - name: Docker meta
-        id: meta
-        uses: docker/metadata-action@v4
-        with:
-          # list of Docker images to use as base name for tags
-          images: |
-            ghcr.io/epinio/epiniod
-          # generate Docker tags based on the following events/attributes
-          tags: |
-            type=ref,event=tag
-            type=sha,enable=true,priority=100,prefix=,suffix=,format=short
+      - uses: chrisdickinson/setup-yq@latest     # Install yq
+      - uses: WyriHaximus/github-action-helm3@v2 # Install helm
+      - name: Docker image tag
+        id: get-tag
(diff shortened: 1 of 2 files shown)


jimmykarily:
Run successfully: https://github.com/epinio/helm-charts/actions/runs/3050841101/jobs/4918392879

jimmykarily:
This PR doesn't fix the fact that the pipeline doesn't trigger automatically upon release. Even if it did, I'm not sure the `helm repo update` would immediately "see" the new version so the `helm  show chart` command may not work.

On the other hand, this PR allows us to run the pipeline manually at any point. This will create an image for the latest released chart, no matter when it was released.

.github/workflows/build-epiniod.yml
```diff
@@ -27,25 +27,21 @@ jobs:
...
+          echo "::set-output name=tag::$latestRelease"
 
       - name: Build and push
         uses: docker/build-push-action@v3
         with:
           push: true
           context: images/.
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
+          tags: ghcr.io/epinio/epiniod:${{ steps.get-tag.outputs.tag }}
```
jimmykarily, line 45: Since this pipeline is always building an image for the latest release, I think it's safe to also push the image with the `latest` tag. I'll add that.

jimmykarily: Done here: 05b427943fb3be38027a7933caa81bde4b4e1c6d

PR #295, opened 2022-10-27
Add Epinio UI Dex client

richard-cox:
- Contributes to https://github.com/epinio/ui/issues/167
- Related PRs https://github.com/rancher/dashboard/pull/7299, https://github.com/epinio/ui-backend/pull/9
- This moves the flag to enable dex out from `dex.enabled` to `global.dex.enabled` as the epinio-ui subchart uses it
- Also refactored epinio-ui chart values to make it clearer which are used directly in the app itself

### DIFF
--- a/chart/epinio/values.yaml
+++ b/chart/epinio/values.yaml
@@ -117,8 +117,6 @@ epinio-ui:
   epinioUI:
     theme: light
     version: "v1.6.1"
-  dex:
-    enabled: true
   ingress:
     enabled: false
 
--- a/chart/epinio/values.yaml
+++ b/chart/epinio/values.yaml
@@ -82,7 +82,7 @@ dex:
     create: false
     name: "dex-config"
   ui:
-    secret: ""
+    secret: "dev-secret"
     # Defaults to https://epinio.{{ .Values.global.domain }}/auth/verify/
     redirectURI: ""
 
--- a/chart/epinio/values.yaml
+++ b/chart/epinio/values.yaml
@@ -75,14 +75,14 @@ api:
 
 # Dex subchart values -- None for now, and sub chart disabled
 dex:
-  enabled: true
   # hardcode this, to avoid problems with release name
   fullnameOverride: "dex"
   configSecret:
     create: false
     name: "dex-config"
   ui:
-    secret: "dev-secret"
+    # secret should be supplied by dex automatically, this is just a fall back
+    secret: ""
     # Defaults to https://epinio.{{ .Values.global.domain }}/auth/verify/
     redirectURI: ""
 
@@ -119,8 +119,6 @@ epinio-ui:
     version: "v1.6.1"
   dex:
     enabled: true
-    ui:
-      secret: "dev-secret"
   ingress:
     enabled: false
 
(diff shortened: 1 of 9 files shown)


richard-cox:
@enrichman There's a couple of dependent PRs to merge before dex will work with the UI, however merging this before they go in should be harmless.

I haven't yet covered the Rancher Desktop issuer/endpoint case, I'd like to get the current set of PRs merged and then make one more targeted if there's an issue

chart/epinio/values.yaml
```diff
@@ -110,8 +114,11 @@ minio:
 
 epinio-ui:
   enabled: true
-  epinioTheme: light
-  epinioVersion: "v1.6.2"
+  epinioUI:
+    theme: light
+    version: "v1.6.1"
+  dex:
+    enabled: true
```
enrichman, line 121: Is this a leftover?

richard-cox: Good spot! Removed

chart/epinio/values.yaml
```diff
@@ -110,8 +114,9 @@ minio:
 
 epinio-ui:
   enabled: true
-  epinioTheme: light
-  epinioVersion: "v1.6.2"
+  epinioUI:
+    theme: light
+    version: "v1.6.1"
```
thehejik, line 119: Wrong version here, it should be `v1.6.2`.

@enrichman given that the key names changed don't we need to adapt our CI somehow to be able do the version bump?

enrichman: Good catch. We need to update also this: https://github.com/epinio/helm-charts/blob/main/updatecli/updatecli.d/epinio.yaml#L89

thehejik: @richard-cox please adapt the key name within this PR here https://github.com/epinio/helm-charts/blob/main/updatecli/updatecli.d/epinio.yaml#L89

thehejik:
Works as expected:
1) Validated with current `epinio/epinio-ui:v1.5.1-0.0.3` (non-oidc) image - still compatible - OK
2) Validated with custom oidc-enabled image - DEX login over UI; DEX login over CLI; non-DEX UI login; deployed an App over UI&CLI - OK

But for both cases I had to update epinio-ui-1.5.3.tgz manually to incorporate my custom image and also to update the ui subchart with current code from this PR.

Not approving yet since this tgz issue needs to be resolved first somehow.

Tested on a single-node `v1.25.6-k3s1` by:
```
helm upgrade --install epinio --namespace epinio ./chart/epinio/ --set skipTraefik=true --set global.domain=REDACTED.nip.io --create-namespace
```

chart/epinio/values.yaml
richard-cox: Version no longer needed and has been removed. It now now tracks whatever version the underlying epinio instance is, and value comes directly from call to `/info`

richard-cox:
@thehejik The sub chart tar is re-created automatically (this last happened via https://github.com/epinio/helm-charts/pull/312/files). I tried to find the job that does this though and couldn't, any ideas?

enrichman (approved):

PR #300, opened 2022-11-14
TEST CI with paketobuilder pinned to 0.2.220-full

andreas-kupries:
that version is 6 days old, when CI was ok.

ref https://github.com/epinio/epinio/issues/1857


andreas-kupries:
Abandoning test work

PR #318, opened 2022-12-14
Fix non use of `server.ingressClassName` by templates

andreas-kupries:
ref: https://github.com/epinio/epinio/issues/1954
epinio import via https://github.com/epinio/epinio/pull/1963

:notes:  this is the ingress class to pass to applications
:memo:  spacing in values.yaml, to enhance readability (separate sections by empty lines)
:hammer: :bug:  typo in same

### DIFF
--- a/chart/epinio/templates/server.yaml
+++ b/chart/epinio/templates/server.yaml
@@ -296,7 +296,10 @@ spec:
             - name: REGISTRY_CERTIFICATE_SECRET
               value: "epinio-registry-tls"
             {{- end }}
-            {{- if .Values.ingress.ingressClassName }}
+            {{- if .Values.server.ingressClassName }}
+            - name: INGRESS_CLASS_NAME
+              value: "{{ .Values.server.ingressClassName }}"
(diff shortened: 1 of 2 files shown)


enrichman:
I'm not a big fan of the mixed uppercase titles, i.e. `Start With An Issue` vs `Start with an issue`. Also in the other pages we have the old format I think. @nunix do we have a general guideline here?

enrichman:
LGTM, I was just wondering if a `server.appIngressClassName` would be more explicit than the `server.ingressClassName`.

andreas-kupries:
Most likely yes. I certainly was confused when I saw this.

andreas-kupries:
It would also be an API change (for operators/installers)

enrichman:
Ah sorry, I thought that it was a new value. I was confused by the Github's diff.

Also if it was not used I would probably try to avoid API changes so.. yep, let's keep it like this. The comment is explaining it, and the `ingressClassName` seems to be pretty standard as a key.

enrichman (approved):

PR #321, opened 2022-12-21
Use static bitnami service-catalog

thehejik:
Changed the bitnami helm chart repo to use the static one `pre-2022`.

Unfortunatelly mongodb is available there in older versions only.


enrichman:
Having a better look at [this issue](https://github.com/bitnami/charts/issues/10539) it seems that the full index.yaml is available here: `https://raw.githubusercontent.com/bitnami/charts/archive-full-index/bitnami`.

PR #324, opened 2022-12-22
s3gw integration

giubacc:
Starting from this patch, Epinio's chart can install s3gw as the S3 storage manager.
To install s3gw as bundle of Epinio, you have to set the flag: `s3gw.enabled` to true.
s3gw is imported as subchart and the related values can be changed under the section s3gw in the `values.yaml` file.
The configuration defaults this way:

  ```
s3gw:
    enabled: false
    ingress:
      enabled: false
    ui:
      enabled: false
    serviceName: s3gw
    useExistingSecret: true
    storageSize: 2Gi
    storageClass:
      create: false
      name: ""
```

For the time being, Minio remains the default option; so s3gw must be explicitly enabled with:

```
minio:
  enabled: false

s3gw:
  enabled: true
```

s3gw uses the default StorageClass to provision its pvc.
The user can specify an explicit StorageClass to use by setting:

```
s3gw:
  storageClass:
    name: "my-storage-class"
```

**please note that we will have to wait for the official 0.11.0 release of s3gw to merge this**

Fixes: https://github.com/epinio/helm-charts/issues/323
Signed-off-by: Giuseppe Baccini <giuseppe.baccini@suse.com>

### CODE REVIEW

**Comment by giubacc** [comment]
Thanks @thehejik for your review/tests!
It seems that this is working reasonably well; btw, I see that linting is failing:
```
Error: Error linting charts: Error building dependencies for chart 'epinio => (version: "1.5.2", path: "chart/epinio")': Error waiting for process: exit status 1
Error: the lock file (Chart.lock) is out of sync with the dependencies file (Chart.yaml). Please update the dependencies 
```

am I required to fix something here?

**Comment by thehejik** [comment]
I guess it is just about running `helm dependency update ./` in `chart/epinio` directory which will update the dependency TGZs and recalculate the digest in Chart.lock file there. Problem is that s3gw version 0.11 wasn't released yet so the command fails. Once it will be released the command should succeed and the lock file will be updated.

File: chart/epinio/values.yaml [hunk 1]
```diff
@@ -53,7 +53,7 @@ certManagerNamespace: cert-manager
 # Connection details for the S3 storage
 s3:
   endpoint: s3.amazonaws.com
-  bucket: ""
+  bucket: "epinio"
```
enrichman, line 56: I would not change the default value for the bucket

enrichman (changes requested):
Nevermind

enrichman (commented):
Just a small comment

chart/epinio/values.yaml
giubacc: I don't agree because the previous logic was still setting `epinio` as default but it was hiding that information inside a non explicit way for the user:
 
`{{- $s3Bucket := .Values.minio.enabled | ternary "epinio" .Values.s3.bucket -}}`

In the new way, the user is clearly seeing what is the default in the values.

thehejik (dismissed):
I tried all the combinations: S3 AWS, minIO with default & custom SC and finally S3GW with default & custom SC.

I've always deployed a sample app using the given object store and in all cases it was working as expected.

Only thing I don't like so much is the name for s3gw PVC called `epinio/epinio-pvc` which is a bit misleading when compared with PVC from minio which are clear `epinio/export-{0..3}-minio-0` (it was always using 4 PVs for minio, no idea why).

Tested on EKS 1.23 with gp2 EBS storage.

thehejik (approved):

PR #341, opened 2023-01-24
Remove the dex cert fighting with the ingress derived dex-tls cert over the dex-tls secret

andreas-kupries:
Ref https://github.com/epinio/epinio/issues/2041

Candidate fix A: Remove the superfluous manually specified cert.

### DIFF
--- a/chart/epinio/templates/dex.yaml
+++ b/chart/epinio/templates/dex.yaml
@@ -40,7 +40,6 @@ stringData:
     - id: epinio-cli
       name: 'Epinio cli'
       public: true
-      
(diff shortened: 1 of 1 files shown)


mmartin24:
@andreas-kupries I was having a look at this and it is very interesting the change before / after on this small video while checking logs on the `cert-manager/cert-manager` with k9s

https://user-images.githubusercontent.com/37271841/214510251-1f423a1a-393a-4939-a459-f0d5a6b697b5.mov

enrichman (approved):

PR #342, opened 2023-01-24
Drop cluster issuer annotation from ingress

andreas-kupries:
Ref: https://github.com/epinio/epinio/issues/2041

Fix candidate B: Remove trigger for auto-generated cert


andreas-kupries:
After discussion we keep auto-gen and remove the manual cert

PR #370, opened 2023-02-20
fix: create nicer-named replacement for `email` template variable

### LINKED ISSUE #348: Missing helm chart value for custom email address used by tlsIssuer
When installing epinio over helm chart user can use `-- set global.tlsIssuer=letsencrypt-production` but there is no way how to specify an email for DNS/cert notifications from the Issuer, it is using hardcoded email `epinio@suse.com` instead.

The hardcoded address is visible for eg. in `kubectl get clusterissuers.cert-manager.io letsencrypt-production -o yaml` output as   `.spec.acme.email` value.

So something like `--set global.tlsIssuer.email=<custom email>` would be nice.

Found in epinio v1.6.1

andreas-kupries:
fix #348 

Note: for backward compat keeping `email` as fallback

Epinio PR ref https://github.com/epinio/epinio/pull/2088
Documentation ref https://github.com/epinio/docs/pull/214

### DIFF
--- a/chart/epinio/values.yaml
+++ b/chart/epinio/values.yaml
@@ -2,7 +2,12 @@
 ## This is a YAML-formatted file.
 ## Declare variables to be passed into your templates.
 
-# The email address you are planning to use for getting notifications about your certificates.
+# Fall back email address to receive notifications from the `letsencrypt-production` issuer.
+#
+# __SUPERCEDED__ by `global.tlsIssuerMail`.
+#
+# Kept for backward compatibility, here and in the templates.
+
 email: "epinio@suse.com"
 
 image:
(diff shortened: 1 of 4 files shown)


chart/epinio/values.yaml
```diff
@@ -174,5 +179,7 @@ global:
   # The URL of the container registry from where to pull container images for the various
   # created Pods. Don't confuse this registry with the "Epinio registry" which is the one
   # where Epinio stores the application images.
+  tlsIssuerMail: "epinio@suse.com"
+  # Email address to receive the certificate notification emails send by the `letsencrypt-production` issuer.
```
enrichman, line 183: I guess that the comment should be right above, keeping the old one to the `cattle`.

I'm also not sure if it should be `tlsIssuerEmail`, instead of `tlsIssuerMail`.

enrichman (changes requested):

chart/epinio/values.yaml
andreas-kupries: ... Rereading ... Right, I somehow thought comments were after the thing
Yes, email is likely better.

enrichman (approved):

PR #371, opened 2023-02-20
make S3 questions consistent in questions.yml

giubacc:
Minio and s3gw checkboxes were not in the same group. Renamed "External S3 storage" into "S3 storage"


andreas-kupries (approved):
Looks ok to me.

PR #372, opened 2023-02-20
s3gw option as subquestion of Minio

giubacc:
Make Install s3gw checkbox as subquestion of Install minio checkbox

### DIFF
--- a/chart/epinio/questions.yml
+++ b/chart/epinio/questions.yml
@@ -114,51 +114,52 @@ questions:
     required: false
 - variable: minio.enabled
   label: Install Minio
-  description: "Disable Minio to use s3gw."
+  description: "Disable Minio to either use s3gw or to configure an external s3 storage."
   type: boolean
   group: "S3 storage"
   show_subquestion_if: false
-- variable: s3gw.enabled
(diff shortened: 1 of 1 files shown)


giubacc:
previous PR was not sufficient: s3gw checkbox must be presented as subquestion of Minio checkbox (showed if disabled)

chart/epinio/questions.yml
```diff
@@ -118,47 +118,48 @@ questions:
   type: boolean
   group: "S3 storage"
   show_subquestion_if: false
-- variable: s3gw.enabled
```
andreas-kupries, line 121: I believe that the minio description should be modified as well.

Instead of   `description: "Disable Minio to use s3gw."`
Use `  description: "Disable Minio to use s3gw or external S3."`
To make it plain at the top-level visible description that this is where external S3 is hidden.

andreas-kupries (changes requested):

chart/epinio/questions.yml
giubacc: done

andreas-kupries (approved):

PR #378, opened 2023-02-22
Fixed epinio-ui deployment with disabled dex

enrichman:
Fix: https://github.com/epinio/ui/issues/176

The `epinio-ui` deployment was failing because of the missing `dex-config` secret when Dex is disabled.

This PR wraps and removes the `EPINIO_DEX_*` environment variables when Dex is not enabled.

It also bumps the `epinio-ui` chart that can be used by the Epinio one.

### DIFF
--- a/chart/epinio-ui/Chart.yaml
+++ b/chart/epinio-ui/Chart.yaml
@@ -15,4 +15,4 @@ name: epinio-ui
 sources:
 - https://github.com/epinio/ui
 type: application
-version: 1.7.0
+version: 1.7.1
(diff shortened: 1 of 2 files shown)


andreas-kupries (approved):
That looks ok. Assuming that the removal of these variables indicates to the UI image/process that dex is not available. If not then the `EPINIO_DEX_ENABLED` variable might have to be moved out of the if/end block and always set as it was.

mmartin24:
Tested on branch and working.

Test steps done:
- Checkout into branch
- Disable dex on `epinio/helm-charts/chart/epinio/values.yaml`
```
global:
  dex:
    enabled: false
```
- On  `epinio/helm-charts/chart/epinio/Chart.yaml` modify the following values:
```yaml
  repository: "file://../epinio-ui"
  tags:
  - epinio-ui
  version: 1.7.1
```
- Execute:
```
helm dependency update helm-charts/chart/epinio
```
- Install epinio: 

Result:

Issue #369, opened 2023-02-20
Move the `epinio-ui` chart to `epinio`


enrichman:
### Is your feature request related to a problem?

Move the `epinio-ui` into the `epinio` chart to simplify the release process.

PR #387, opened 2023-03-07
Removed `epinio-ui` chart dependency

enrichman:
Fix: https://github.com/epinio/helm-charts/issues/369

This PR removes the `epinio-ui` chart as a dependency of `epinio` and brings into the `epinio` chart the needed file to deply the UI.


chart/epinio/templates/ui/deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: epinio-ui
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
+      containers:
+      - name: epinio-ui
+        image: "{{ default .Values.image.epinio-ui.registry (include "registry-url" .) }}{{ .Values.image.epinio-ui.repository }}:{{ .Values.image.epinio-ui.tag }}"
```
andreas-kupries, line 25: linter: `Error:  templates/: parse error at (epinio/templates/ui/deployment.yaml:25): bad character U+002D '-'`
The `epinio-ui` ? Might treat it as subtraction symbol, not part of the name.

enrichman: Ok, now it should be fixed! :laughing:

andreas-kupries (approved):

PR #391, opened 2023-03-13
Allow to override default application port

giubacc:
### Specify a custom listening port for an application (#1792)
User can now specify a custom listening port for an Epinio application.
The port is pushed inside the application container via the environment
variable: `PORT`.
The value defaults to 8080; it can be specified using a custom chart value
during the application's push:

`epinio push -n 'app-name' --chart-value 'appListeningPort=8087'`

Valid appListeningPort values range from `1001` to `(2^16)-1`

Fixes: https://github.com/epinio/epinio/issues/1792

### DIFF
--- a/chart/application-stateful/Chart.yaml
+++ b/chart/application-stateful/Chart.yaml
@@ -4,7 +4,7 @@ description: The helm chart for epinio applications to be deployed by
 icon: https://raw.githubusercontent.com/epinio/helm-charts/main/assets/epinio.png
 home: https://github.com/epinio/epinio
 type: application
-version: 0.1.21
+version: 0.1.22-rc1
 keywords:
 - epinio
 - paas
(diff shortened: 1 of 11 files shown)


andreas-kupries (changes requested):
IIRC it is necessary to bump the version number in the application chart (i.e. 24 -> 25), and make the same change in the epinio chart to ensure that we have a proper new release of the standard chart with the new feature.

I also believe that we should have an epinio PR as well which
  - imports the changed epinio chart, and
  - has a test demonstrating the passing of the new port, to an application which reacts to that port.

While we have tests demonstrating the CVs in general I believe that this new feature does need an explicit test, given that our standard chart changes.

Note: The golang sample app in our assets looks to recognize and use the `PORT` environment variable.

> assets/golang-sample-app/main.go:36:	log.Fatal(http.ListenAndServe(":"+os.Getenv("PORT"), nil))

giubacc:
thanks @andreas-kupries , I've bumped application charts versions; what is now the immediate next step todo?

andreas-kupries (approved):

PR #395, opened 2023-03-22
Bump application chart versions

giubacc:
- application chart: `0.1.25`
- stateful chart: `0.1.22`

Addresses: https://github.com/epinio/epinio/issues/1792
Signed-off-by: Giuseppe Baccini <giuseppe.baccini@suse.com>

### DIFF
--- a/chart/application-stateful/Chart.yaml
+++ b/chart/application-stateful/Chart.yaml
@@ -4,7 +4,7 @@ description: The helm chart for epinio applications to be deployed by
 icon: https://raw.githubusercontent.com/epinio/helm-charts/main/assets/epinio.png
 home: https://github.com/epinio/epinio
 type: application
-version: 0.1.22-rc1
+version: 0.1.22
 keywords:
 - epinio
 - paas
--- a/chart/application/Chart.yaml
+++ b/chart/application/Chart.yaml
@@ -4,7 +4,7 @@ description: The helm chart for epinio applications to be deployed by
 icon: https://raw.githubusercontent.com/epinio/helm-charts/main/assets/epinio.png
 home: https://github.com/epinio/epinio
 type: application
-version: 0.1.25-rc1
+version: 0.1.25
 keywords:
 - epinio
 - paas


chart/epinio/templates/default-app-chart.yaml
```diff
@@ -12,4 +12,7 @@ metadata:
 spec:
   shortDescription: Epinio standard deployment
   description: Epinio standard support chart for application deployment
-  helmChart: https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.24/epinio-application-0.1.24.tgz
+  helmChart: https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.25/epinio-application-0.1.25.tgz
```
olblak, line 15:
```suggestion
  helmChart: https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.25/epinio-application-0.1.25.tgz
```
Is the version specified in this url the same one than the one specificied in the Chart.yaml? 
If yes, then it would be better to use templating.

Hardcoded version in template files are not easy to find

giubacc: I have restored the old code, as @enrichman said, this specific update of pointed version should be handled automatically by an automatic PR.

olblak: Ha ok sorry

enrichman: @giubacc I was not probably clear that we need to wait the update also to add the settings part, or the test will fail.
Basically we cannot touch the Epinio chart until we have the new release ready.

chart/epinio/templates/default-app-chart.yaml
```diff
@@ -13,3 +13,6 @@ spec:
   shortDescription: Epinio standard deployment
   description: Epinio standard support chart for application deployment
   helmChart: https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.24/epinio-application-0.1.24.tgz
+  settings:
+    appListeningPort:
+      type: 'integer'
```
enrichman, line 18:
```suggestion
```

enrichman (approved):

PR #419, opened 2023-04-26
feat: updated app CRD in chart

andreas-kupries:
ref https://github.com/epinio/epinio/issues/2237

### DIFF
--- a/chart/epinio/crds/app-crd.yaml
+++ b/chart/epinio/crds/app-crd.yaml
@@ -64,10 +64,16 @@ spec:
                 type: string
               origin:
                 properties:
+                  archive:
+                    type: boolean
                   container:
                     type: string
                   git:
                     properties:
+                      branch:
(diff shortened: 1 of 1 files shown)


enrichman (commented):
I'm not sure about the versioning. Will the helm upgrade work If the CRDs has the same `v1` version?

andreas-kupries:
> I'm not sure about the versioning. Will the helm upgrade work If the CRDs has the same `v1` version?

I believe so. IIRC we had no issues with past extensions of the CRD, and we never bumped v1 to something else.
Maybe because that works, if you only extend the CRD, and not remove/rename things.

enrichman:
Helm will not update the CRD, but we don't have other options. I think we can do some manual check trying a couple of manual things and see what is going to happen.

Those parameters are not required though, so it shouldn't be an issue anyway.

LGTM :+1:

PR #434, opened 2023-05-30
Added `containerregistry.certificateSecret` value

enrichman:
This PR adds the `containerregistry.certificateSecret` value.

This can be used to specify the secret containing the certificate of an external registry, that needs to be trusted by the staging job.

### DIFF
--- a/chart/epinio/templates/server.yaml
+++ b/chart/epinio/templates/server.yaml
@@ -309,6 +309,9 @@ spec:
             {{- if .Values.containerregistry.enabled }}
             - name: REGISTRY_CERTIFICATE_SECRET
               value: "epinio-registry-tls"
+            {{- else if .Values.containerregistry.certificateSecret  }}
+            - name: REGISTRY_CERTIFICATE_SECRET
+              value: {{ .Values.containerregistry.certificateSecret }}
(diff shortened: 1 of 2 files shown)


thehejik (approved):
LGTM and tested as follows:
```bash
$ kubectl create ns epinio
$ kubectl create secret -n epinio generic epinio-external-registry-tls --from-file=tls.crt=/home/sles/certs/subCA.pem
$ helm upgrade --install epinio --namespace epinio ./chart/epinio/ \
    --set global.domain=redacted.io \
    --set global.registryURL=redacted.dev \
    --set global.registryUsername=opensuse \
    --set global.registryPassword=password \
    --set containerregistry.enabled=false \
    --set image.epinio.tag=v1.9.0-rc2 \
    --set containerregistry.certificateSecret=epinio-external-registry-tls
```

The secret is set in epinio-server deployment:
```bash
$ kubectl get deploy -n epinio epinio-server  -o jsonpath='{..env}' | jq
...
  {
    "name": "REGISTRY_CERTIFICATE_SECRET",
    "value": "epinio-external-registry-tls"
  }
```

Also an app deployment using the registry tls/secret is possible

PR #435, opened 2023-05-30
Added `strategy` to Epinio deployment

enrichman:
This PR adds an optional `strategy` values that can be used to decide the deployment strategy of the Epinio server.

Using a `maxSurge=0` and a maxUnavailable=1` will fix the `MultiAttach` error while using some RWO storage.

Fixes https://github.com/epinio/epinio/issues/2253

### DIFF
--- a/chart/epinio/templates/server.yaml
+++ b/chart/epinio/templates/server.yaml
@@ -246,6 +246,10 @@ spec:
       app.kubernetes.io/instance: default
       app.kubernetes.io/name: epinio-server
       app.kubernetes.io/part-of: epinio
+  {{- with .Values.strategy }}
+  strategy:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   template:
     metadata:
       labels:
--- a/chart/epinio/values.yaml
+++ b/chart/epinio/values.yaml
@@ -64,6 +64,15 @@ service:
   # -- Annotations to be added to the Epinio service.
   annotations: {}
 
+# The strategy used to deploy the Epinio server.
+# If you are using a RWO storage the following will avoid a Multi-Attach error during an `helm upgrade`.
+# See https://github.com/epinio/epinio/issues/2253.
+strategy:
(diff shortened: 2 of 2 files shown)


thehejik (approved):
Fix works as expected. Tested on 2 node RKE2 with RWO Longhorn SC.

I was able to reproduce the problem with upstream chart for Epinio v1.8.1.

Then I did `helm upgrade epinio` using this PR tree and `--set image.epinio.tag=v1.9.0-rc2` flag:
* I verified that the strategy is applied by:
```
$ kubectl get deployment -n epinio epinio-server -o jsonpath='{..strategy}' | jq
{
  "rollingUpdate": {
	"maxSurge": 0,
	"maxUnavailable": 1
  },
  "type": "RollingUpdate"
}
```
* the existing server pod (deployed by upstream chart) has been terminated immediately and the new server pod started on the second node without the multi-attach PV error

I've also verified whether minio and s3gw are affected by the same issue but they are not:
- **minio** uses STS deployment so it will basically keep the pod running on the node where it was running before the upgrade (if the minio tag changes) - if the `minio-0` pod will be, for whatever reason as drain node, scheduled on the second node the PV will be mounted (or recreated when using `local-path`) there as well
- **s3gw** uses `strategy.type=Recreate` which always deletes existing RS and terminates its pods immediately, then new RS is created and new pods are scheduled. It is actually pretty similar to the approach used in this PR but it might take longer to run the pod.

NOTE commands:
```
$ helm upgrade --install epinio --namespace epinio ./chart/epinio --set global.domain=10.100.103.61.nip.io --create-namespace --set image.epinio.tag=v1.9.0-rc2
$ kubectl get pods -n epinio -o custom-columns=NAME:.metadata.name,NODE:.spec.nodeName,STATUS:..phase
$ kubectl drain <node> --ignore-daemonsets --delete-emptydir-data # for pod eviction
$ kubectl uncordon <node> # to make the node Ready again
```

PR #439, opened 2023-06-16
Expose staging job resource requests in chart to user

andreas-kupries:
ref https://github.com/epinio/epinio/issues/2286

feat: Expose resource requests for staging job to chart users
chore: update schema definition
fix: schema definition (added some missing older things)
chore: pass resource requests to server, per environment variables


thehejik (approved):
Tested [here](https://github.com/epinio/epinio/pull/2384#pullrequestreview-1486192563)

PR #466, opened 2023-07-18
fix S3_CERTIFICATE_SECRET env var value when using s3gw

### LINKED ISSUE #470: Bump s3gw dependency

giubacc:
Fix https://github.com/epinio/epinio/issues/2466
Fix #470 - Added

(andreas-kupries) My plan is to have fix and bump (back to 0.18) in a single PR, this here.


thehejik:
The app deployment when using s3gw 0.18.0 works ok with this fix. Anyway we downgraded to s3gw 0.14.0 where this fix is not needed (and has correct resource names) so this might be relevant for later E releases with bumped s3gw.

andreas-kupries (approved):

giubacc:
this PR must be patched using `-s3gw-cluster-ip-tls` as secret name; this can be done only when we bump the s3gw to `0.21.0`; not yet released.

PR #467, opened 2023-07-18
Revert `s3gw` to `0.14.0`

enrichman:
Since we are close to the release and we have faced some problems during the upgrade of the `s3gw` we would like to have more time to investigate those issues, and provide a clean upgrade release note.

This revert the `s3gw` to the version `0.14.0` (https://github.com/epinio/helm-charts/pull/459/files) ~and remove the fix introduced in https://github.com/epinio/helm-charts/pull/463.~

For the update we also need to add this: https://github.com/epinio/helm-charts/pull/466

/cc @thehejik @giubacc


chart/epinio/values.yaml
```diff
@@ -169,7 +169,6 @@ s3gw:
     enabled: false
   serviceName: s3gw
   useExistingSecret: true
-  defaultUserCredentialsSecret: s3gw-creds
```
giubacc, line 172: I would leave this, no need to remove, when we will retry a bump we will already have this.

giubacc (changes requested):

thehejik (approved):
Tested incl. sample app deployment using s3gw store, all good.

giubacc (approved):

thehejik:
Verified fixed in epinio v1.9.0-rc5

PR #478, opened 2023-08-07
added support for configurable size of staging disk.

andreas-kupries:
ref https://github.com/epinio/epinio/issues/2504
ref


thehejik (approved):
Tested and verified in https://github.com/epinio/epinio/pull/2508#pullrequestreview-1567281881

PR #490, opened 2023-09-11
Added 'adminPassword', 'epinioPassword', and 'password' values

enrichman:
This PR adds the 'api.adminPassword', 'api.epinioPassword' fields to specifiy the cleartext password for the `admin` and `epinio` user. This password will be then hashed with Bcrypt.

In the same way now also a `password` field was added in the `api.users` object.

If the `adminPassword` or the `epinioPassword` are specified they will have priority over the `passwordBcrypt` or `password` fields. Otherwise the `passwordBcrypt` has priority over the `password` field.

In the Rancher `question.yaml` the `api.username` or the `api.passwordBcrypt` fields were not working, probably a leftover of the initial fixed user to the users array refactor. Now the `adminPassword` field can be specified, but we don't have a way to provide the username of the defaul/admin user.

### DIFF
--- a/chart/epinio/questions.yml
+++ b/chart/epinio/questions.yml
@@ -56,15 +56,9 @@ questions:
     - "epinio-ca"
     - "selfsigned-issuer"
     - "letsencrypt-production"
-- variable: api.username
-  label: API username
-  description: "The user name for authenticating all API requests"
-  type: string
-  required: false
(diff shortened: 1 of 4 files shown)


andreas-kupries (approved):
Ok. Despite the hackishness. I do not see a way around it either.

PR #495, opened 2023-09-13
Install `letsencrypt-[staging|production]` ClusterIssuer only when required

### LINKED ISSUE #479: Allow to disable "letsencrypt-production" cluster-issuer installation
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Today when we install the Epinio helm chart, we also install the "letsencrypt-production" clusterissuer by default
While it is useful for development environment, it's problematic in the following scenarios:

https://github.com/epinio/helm-charts/blob/a558472d70279bb9e417a5b19143def4a2bf3f7a/chart/epinio/templates/cluster-issuers.yaml#L15

* The `letsencrypt-production` clusterissuer was already installed previously, this result in a failure when installing the Epinio helm chart

* The `letsencrypt-production` clusterissuer comes with default email set to `epinio@suse.com` 

### Expected Behavior

I would like to opt-out by default 

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- Architecture:
- Epinio Version:
- Cluster:
  - Provider:
  - Options:
  - Kubernetes Version:
```

### Logs

_No response_

### Anything else?

_No response_

enrichman:
Fix #479 

This PR avoid the installation of the `letsencrypt-production` when not needed. It will be created only when the tlsIssuer value is set. It also adds the staging issuer, and fixes a couple of wrong questions values during the Rancher installation.

### DIFF
--- a/chart/epinio/values.yaml
+++ b/chart/epinio/values.yaml
@@ -215,7 +215,7 @@ global:
   # Used in registry path when pushing -> "external.tld/apps/APPNAME"
   registryNamespace: "apps"
   # The name of the cluster issuer to use.
-  # Epinio creates three options: 'epinio-ca', 'letsencrypt-production', and 'selfsigned-issuer'.
+  # Epinio creates three options: 'epinio-ca', 'selfsigned-issuer', 'letsencrypt-staging' and 'letsencrypt-production'.
   tlsIssuer: "epinio-ca"
   # The name of your ClusterIssuer (it will override the tlsIssuer)
   customTlsIssuer: ""
--- a/chart/epinio/values.yaml
+++ b/chart/epinio/values.yaml
@@ -217,8 +217,10 @@ global:
   # The name of the cluster issuer to use.
   # Epinio creates three options: 'epinio-ca', 'letsencrypt-production', and 'selfsigned-issuer'.
   tlsIssuer: "epinio-ca"
+  # The name of your ClusterIssuer (it will override the tlsIssuer)
+  customTlsIssuer: ""
   # Email address to receive the certificate notification emails send by the `letsencrypt-production` issuer.
-  tlsIssuerEmail: "epinio@suse.com"
+  tlsIssuerEmail: ""
   # The URL of the container registry from where to pull container images for the various
   # created Pods. Don't confuse this registry with the "Epinio registry" which is the one
   # where Epinio stores the application images.
--- a/chart/epinio/questions.yml
+++ b/chart/epinio/questions.yml
@@ -55,6 +55,7 @@ questions:
   options:
     - "epinio-ca"
     - "selfsigned-issuer"
+    - "letsencrypt-staging"
     - "letsencrypt-production"
 - variable: api.adminPassword
   label: API password for the 'admin' user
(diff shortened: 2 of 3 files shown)


chart/epinio/values.yaml
```diff
@@ -215,10 +215,12 @@ global:
   # Used in registry path when pushing -> "external.tld/apps/APPNAME"
   registryNamespace: "apps"
   # The name of the cluster issuer to use.
-  # Epinio creates three options: 'epinio-ca', 'letsencrypt-production', and 'selfsigned-issuer'.
+  # Epinio creates three options: 'epinio-ca', 'selfsigned-issuer', 'letsencrypt-staging' and 'letsencrypt-production'.
```
thehejik, line 218:
```suggestion
  # Epinio creates four options: 'epinio-ca', 'selfsigned-issuer', 'letsencrypt-staging' and 'letsencrypt-production'.
```

thehejik:
### Test

Environment: RKE2 on EC2 with aws cloud provider installed, local-path storage provisioner, nginx-ingress-controller and a wildcard enabled public FQDN domain from zone managed by aws route53.

We always installed Epinio from this branch and then deployed a sample app - then verified all the endpoints (epinio, auth, app) if these are signed by corresponding TLS cert.

#### Incorporated  `ClusterIssuer`s:
* **OK** `--set global.tlsIssuer=letsencrypt-staging --set global.tlsIssuerEmail=email@suse.com`
* **OK** `--set global.tlsIssuer=letsencrypt-production --set global.tlsIssuerEmail=email@suse.com`

Verified that either `production` or `staging` clusterIssuers are created only when requested by the helm flags above.

#### Custom/existing  `ClusterIssuer`: 
* **NOK** `--set global.customTlsIssuer=letsencrypt-production` - we checked the custom issuer is not overwritten by the one from epinio and it works fine for `epinio` and `auth` endpoints but doesn't work for deployed application which always uses `epinio-ca` issuer instead.



Note1: the used issuers can be verified in browser or by command `k get certificaterequests.cert-manager.io  -A` command
Note2: custom `ClusterIssuer` can be defined by something like:
```bash
kubectl apply -f - <<EOF
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: letsencrypt-production
  namespace: cert-manager
  annotations:
    customfield: my-custom-issuer
spec:
  acme:
    server: https://acme-v02.api.letsencrypt.org/directory
    email: email@suse.com
    privateKeySecretRef:
      name: letsencrypt-production
    solvers:
    - http01:
        ingress:
          class: nginx
EOF
```

andreas-kupries (approved):
Note for the release notes:
- Installation change: "Does not install the letsencrypt issuers by default any more".

chart/epinio/values.yaml
enrichman: Hm, yes. But it's not really clear from the comment. Not sure how to rephrase it.

enrichman:
> Note for the release notes:
> 
> * Installation change: "Does not install the letsencrypt issuers by default any more".
> 
> Should also explain the various configurations (no issuer, issuer, custom issuer, special values) in the docs.

Added some more comments here: https://github.com/epinio/docs/pull/312

enrichman:
@thehejik awesome! Thanks for the further check on Rancher. :cow2:

PR #504, opened 2023-10-18
Optionally supplement Dex configuration with information required for embedded ui auth

richard-cox:
- Required by https://github.com/epinio/ui/pull/311
- Epinio UI, hosted in Rancher, will need to auth via dex
- To support this the dex config needs updating
   - Allow the rancher domain to contact Dex
   - Add a new client specifically for requests from rancher's ui
- These changes are only applied via new chart value `rancher.url`
  - This has been added to questions.yaml


chart/epinio/templates/dex.yaml
```diff
@@ -102,4 +116,4 @@ spec:
     - "auth.{{ .Values.global.domain }}"
     secretName: dex-tls
 
-{{- end }}
+{{- end }}
```
andreas-kupries, line 119: missing trailing end-of-line

andreas-kupries (changes requested):
one nit (missing end-of-line at end of file).
one big question (about a possible mis-committed file)
the rest looks to be ok.

chart/epinio/get_helm.sh
```diff
@@ -0,0 +1,331 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
richard-cox: Not sure how that crept in! Removed

chart/epinio/templates/dex.yaml
richard-cox: Fixed

andreas-kupries (approved):
It looks ok to me.
I would like to have approval from @enrichman also, as the main dev for the dex/oidc parts.

enrichman (dismissed):
Yes, it looks good to me.

The Dex config update is probably something that we need to test.
I'm not sure about the behavior across Epinio upgrades.

/cc @mmartin24 @thehejik

richard-cox:
I had a think, from memory helm managed resources aren't something that the user should be manually changing. We show a warning in the rancher ui for resources managed by helm saying something similar. So I'm guessing any changes made by the user will get overwritten on upgrade.

chart/epinio/questions.yml
```diff
@@ -173,3 +173,9 @@ questions:
   required: false
   type: string
   group: "Debugging"
+- variable: rancher.url
```
thehejik, line 176: Would be possible to replace the existing one `server.accessControlAllowOrigin` by this new? 
```
server:
  # Domain which serves the Rancher UI (to access the API)
  accessControlAllowOrigin: ""
```

richard-cox: Done. `server.accessControlAllowOrigin` has been replaced with `server.url`.

@enrichman Can we add something that might stand out in release notes, to ensure people who upgrade can switch over?

thehejik:
It works as expected and the `dex-config` has the same fields related to `rancher-dashboard` as in report above.  Also `questions.yml` modification is working properly (@mmartin24 the CORS input field has been moved from `General settings` under a new `Rancher` section):
image

@richard-cox But today I found out that when the chart is installed over rancher:v2.8.0-rc3, there is already value `global.cattle.rancher.url` value automagically set by rancher so maybe we could just reuse the value and remove `rancher.url` completely, wdyt?

thehejik (approved):
Test report:
* OK - deployed out of rancher without `rancher.url` set - the secret data NOT modified
* OK - deployed out if rancher with `rancher.url` set - the secret data modified

* OK - deployed in rancher - the CORS field not present anymore - the secret data modified
* OK - prev. released E version deployed in rancher -> upgrade from CLI by using the values.yaml from rancher installation extended by `rancher.url`  - the secret data modified
* OK - prev. released E version deployed in rancher -> upgrade from CLI by using the values.yaml from rancher installation without `rancher.url` - the secret data modified
* OK - prev. released E version deployed in rancher -> upgrade from CLI by using the values.yaml from rancher installation without `rancher.url` and `global.cattle.url` - the secret data NOT modified
* OK - upgraded over CLI to the chart under test with removed `global.cattle.url` - the secret data doesn't contain the `rancher-dashboard` client anymore

All good including questions.yml, the CORS field is gone but the URL is rendered correctly when installed over rancher.

richard-cox:
> OK - deployed in rancher - the CORS field not present anymore - the secret data modified
@thehejik Just to confirm... in this test case the `ACCESS_CONTROL_ALLOW_ORIGIN` value in the `epinio-server` Deployment is populated?

thehejik:
> @thehejik Just to confirm... in this test case the `ACCESS_CONTROL_ALLOW_ORIGIN` value in the `epinio-server` Deployment is populated?

Yes, it is:
image

richard-cox:
@andreas-kupries , @enrichman unless there are objections i'll merge this tomorrow AM

enrichman:
I will be very happy so see this merged! Thanks @richard-cox

PR #511, opened 2023-11-14
dropped epinio-ui cert, and associated epinio-ui-tls secret

andreas-kupries:
ui ingress now uses the epinio-tls secret providedb y the epinio cert.


enrichman (approved):

thehejik (approved):
Tested with (default `epinio-ca` clusterIssuer only) and without CM inc. app deployment, all good

PR #523, opened 2023-11-21
detect single directory as app source and lift contents to match paketo's expectations

andreas-kupries:
ref https://github.com/epinio/epinio/issues/2465
main PR https://github.com/epinio/epinio/pull/2712

### DIFF
--- a/chart/epinio/templates/stage-scripts.yaml
+++ b/chart/epinio/templates/stage-scripts.yaml
@@ -63,16 +63,13 @@ data:
        ( echo Txz? ; tar -xvJf "../${BLOBID}" ) || \
        ( echo "Unable to unpack. No supported archive file format found" ; exit 1 )
        echo OK
-       if test $(ls|wc -l) == 1 ; then
-         # archive contains single element ...
-         if test -d "$(ls)" 2>/dev/null ; then
-           # ... which is a directory. assume that the actual app sources are
-           # inside it, and lift them out to match paketo's expectations.
-           child="$(ls)"
-           echo Lifting \"${child}\" contents ...
-           mv -v "${child}"/* "${child}"/.??* .
-           rmdir -v "${child}"
-         fi
+       if test $(find . -maxdepth 1 -mindepth 1 -printf %y)  = "d" ; then
+         # archive contains single element which is a directory. assume that the actual
+         # app sources are inside it, and lift them out to match paketo's expectations.
+         child="$(ls)"
+         echo Lifting \"${child}\" contents ...
+         mv -v "${child}"/* "${child}"/.??* .
+         rmdir -v "${child}"
        fi
     )
     rm "/workspace/source/${BLOBID}"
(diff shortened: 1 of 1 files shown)


chart/epinio/templates/stage-scripts.yaml
```diff
@@ -63,6 +63,17 @@ data:
...
+         # archive contains single element ...
+         if test -d "$(ls)" 2>/dev/null ; then
+           # ... which is a directory. assume that the actual app sources are
+           # inside it, and lift them out to match paketo's expectations.
+           child="$(ls)"
+           echo Lifting \"${child}\" contents ...
+           mv -v "${child}"/* "${child}"/.??* .
+           rmdir -v "${child}"
+         fi
+       fi
```
enrichman, line 76: You can use `find` to check if there is only one directory:

- `-maxdepth 1` will check only the current folder
- `-mindepth 1` will exclude the `.`
- `-printf %y` will print the file type (`d` for)


```
find . -maxdepth 1 -mindepth 1 -printf %y
```

Example:

```
mkdir mydir
```
```
ls -al     
total 0
drwxr-xr-x 1 enrico enrico   10 Nov 22 13:52 .
drwxr-xr-x 1 enrico enrico 3900 Nov 22 13:53 ..
drwxr-xr-x 1 enrico enrico    0 Nov 22 13:52 mydir
```
```
find . -maxdepth 1 -mindepth 1 -printf %y
d
```

```suggestion
       # archive contains a single directory
       if [ "$(find . -maxdepth 1 -mindepth 1 -printf %y)" = "d" ] ; then
         # assume that the actual app sources are
         # inside it, and lift them out to match paketo's expectations.
         child="$(ls)"
         echo Lifting \"${child}\" contents ...
         mv -v "${child}"/* "${child}"/.??* .
         rmdir -v "${child}"
       fi
```

enrichman (commented):
Small suggestion using `find`

chart/epinio/templates/stage-scripts.yaml
andreas-kupries: Done. Commit bf615fc8b7c92dcb6c744f09c6214a817d855935

enrichman (approved):

chart/epinio/templates/stage-scripts.yaml
andreas-kupries: Something in my implementation of this is broken. Fails to run.

andreas-kupries: `unpack-blob CHECK=dd` ... Expected `CHECK=d` ... 

Ah: 
```
 unpack-blob .php.ini.d/ 
 unpack-blob .php.ini.d/extensions.ini 
 unpack-blob wordpress/ 
```

A hidden directory ignored by my `ls`, and not ignored the new `find`.
The fix is to undo the change to the text, i.e. go back to `BP_PHP_WEB_DIR=wordpress`.

andreas-kupries: Done so in https://github.com/epinio/epinio/pull/2712. No changes here.

PR #524, opened 2023-11-22
fix: handling of EVs in staging, when no EVs are present at all

andreas-kupries:
ref https://github.com/epinio/epinio/issues/2715
ref https://github.com/epinio/epinio/pull/2717

### DIFF
--- a/chart/epinio/templates/stage-scripts.yaml
+++ b/chart/epinio/templates/stage-scripts.yaml
@@ -74,7 +74,7 @@ data:
     )
     rm "/workspace/source/${BLOBID}"
     mkdir -p /workspace/source/env
-    cp -vL /workspace/source/appenv/* /workspace/source/env
+    cp -vL /workspace/source/appenv/* /workspace/source/env 2>/dev/null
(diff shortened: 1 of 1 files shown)


chart/epinio/templates/stage-scripts.yaml
```diff
@@ -95,7 +95,7 @@ data:
       exit 1
     fi
     find /workspace
-    for path in /workspace/source/env/* ; do export "$(basename "${path}")"="$(cat "${path}")" ; done
+    for path in $(ls /workspace/source/env/* 2>/dev/null) ; do export "$(basename "${path}")"="$(cat "${path}")" ; done
```
enrichman, line 98: Would `find` work here? Not a bash guru but maybe it nicer.

andreas-kupries: `find` would also see the internal `..*` directories and links in the env directory. While constraining it should be possible, that also ups the complexity. IOW might not be as nice as you believe at the moment.

enrichman (approved):

PR #608, opened 2025-10-09
Add image block to each service in Helm chart

### LINKED ISSUE #607: Image Tags Need to Be Updated on Epinio Services - Bitnami Breaking Change
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Because of changes made with [bitnami](https://thenewstack.io/broadcom-ends-free-bitnami-images-forcing-users-to-find-alternatives/)

the epinio services that are using the bitnami charts in the [service catalog](https://github.com/epinio/helm-charts/blob/main/chart/epinio/templates/service-catalog.yaml) don't work. So basically none of the out of the box ones are working at the moment. 

Image

Image

### Expected Behavior

it should deploy the helm charts successfully

### Steps To Reproduce

- setup kind cluster with cert-manager and ingress-nginx. here is how I currently do this [link to gist](https://gist.githubusercontent.com/josephaw1022/09330c8e3b3eaabef6de192e8cdcc93b/raw/7bd6c3e4282a338d706b942365584a8a385dd4bf/kind-dev-cluster.sh)

- then install epinio onto kind cluster

```
cat <<'EOF' | helm upgrade --install epinio epinio/epinio \
  --namespace epinio \
  --create-namespace \
  --values -
global:
  domain: localhost
ingress:
  annotations:
    cert-manager.io/cluster-issuer: quick-kind-ca
EOF

```

### Environment

```markdown
- OS: Linux
- Architecture: x86_64
- Epinio Version: 1.13.0
- Cluster:
  - Provider: KinD
  - Options: 1 master, 1 worker, rancher.io/local-path, no lb just ingress-nginx -> node port
  - Kubernetes Version: 1.34.0
```

### Logs

```markdown

```

### Anything else?

# To Fix Issue

I would first try and update the images used by the charts to a non bitnami image. honestly the official ones on dockerhub or from suse registry (they dont have rabbitmq, mysql, 

run something along the line of these commands to find where the images are in the values file for each chart

```
helm show values bitnami/postgresql --version 12.1.6 | grep -m1 -A 10 -B 10 -E '^[[:space:]]*image:'
helm show values bitnami/mysql --version 9.4.5 | grep -m1 -A 10 -B 10 -E '^[[:space:]]*image:'
helm show values bitnami/redis --version 17.3.17 | grep -m1 -A 10 -B 10 -E '^[[:space:]]*image:'
helm show values bitnami/rabbitmq --version 11.2.2 | grep -m1 -A 10 -B 10 -E '^[[:space:]]*image:'
helm show values bitnami/mongodb --version 13.6.2 | grep -m1 -A 10 -B 10 -E '^[[:space:]]*image:'
```

here are links to the dockerhub repos for convenience

[postgres](https://hub.docker.com/_/postgres)
[rabbitmq](https://hub.docker.com/_/rabbitmq)
[mysql](https://hub.docker.com/_/mysql)
[redis](https://hub.docker.com/r/valkey/valkey)
[mongo](https://hub.docker.com/_/mongo)

and if after updating the images, it works, then sweet! problem fixed 😎. 

BUT...., if some unexpected issue(s) come up with bitnami chart and non-bitnami images, then I would look at changing chart used. But I wouldn't do all that initially, unless it's actually needed.

josephaw1022:
closes #607

### DIFF
--- a/chart/epinio/templates/service-catalog.yaml
+++ b/chart/epinio/templates/service-catalog.yaml
@@ -1,4 +1,3 @@
-# These are three simple Services to fill the Service Catalog
 {{ if .Values.serviceCatalog.enableDevServices }}
 ---
 apiVersion: application.epinio.io/v1
@@ -23,6 +22,12 @@ spec:
     url: "https://charts.bitnami.com/bitnami"
   values: |-
     {{- template "epinio.catalog-service-values" . }}
+    image:
+      registry: docker.io
+      repository: postgres
+      tag: "18.0"
+      digest: ""
+      pullPolicy: IfNotPresent
 ---
 apiVersion: application.epinio.io/v1
 kind: Service
@@ -46,6 +51,12 @@ spec:
     url: "https://charts.bitnami.com/bitnami"
   values: |-
     {{- template "epinio.catalog-service-values" . }}
+    image:
+      registry: docker.io
+      repository: mysql
+      tag: "9.4.0"
+      digest: ""
+      pullPolicy: IfNotPresent
 ---
 apiVersion: application.epinio.io/v1
 kind: Service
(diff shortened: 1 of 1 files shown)


krumware (changes requested):
Should we use this instead of `repository: bitnamilegacy/{image}`?

Also, we can't misrepresent redis by using the valkey image, but I'm happy to accept an addition of a valkey service manifest here!

dcharles525:
Stale, closing.

PR #655, opened 2026-02-27
Epnio 467 use pack

psanghvi17:


chart/epinio/values.yaml
```diff
@@ -38,6 +38,10 @@ image:
     registry: quay.io/
     repository: skopeo/stable
     tag: v1.13.2
+  pack:
+    registry: ""
+    repository: buildpacksio/pack
+    tag: 0.38.2
```
grudra7714, line 44: any reason why we are using 0.38? Latest pack version is 0.40

grudra7714 (commented):
still reviewing

chart/epinio/values.yaml
grudra7714: https://github.com/buildpacks/pack/releases

PR #663, opened 2026-03-16
chore: add kubeVersion constraint to epinio Chart.yaml

cResults:
Add kubeVersion: '>=1.21-0' to Chart.yaml so the chart displays correctly in GKE-based Rancher environments and satisfies the rancher/partner-charts submission requirement that charts have kubeVersion set in metadata.


dcharles525 (dismissed):
LGTM

PR #668, opened 2026-04-13
cleanup and organization of resources

hadams95:
- Add optional cert-manager install flag to allow simplification of install step
- Create directories and split resources out into respective files by subject
    - Remove monolithic files with multiple conditions where possible
    - Reduce chart's root level file sprawl
- Remove `epinio-ui` directory since `ui` is baked into the `epinio` chart


dcharles525 (approved):
LGTM! Thanks!

PR #694, opened 2026-07-15
Fix/reflector failure

dcharles525:
Bumps reflector to fix race condition bug causing the container to stop copying namespace secrets.


psanghvi17:
LGTM

